PR #93, opened 2021-12-02
Handle pagination on subscan transactions

maycon-mello:


src/features/transactions/transactions-slice.js
```diff
@@ -55,45 +56,63 @@ export const transactionsSelectors = {
 };
 
 export const transactionsOperations = {
+  loadExternalTransactions: account => async (dispatch, getState) => {
+    const realm = getRealm();
+    const dbTransactions = realm.objects('Transaction').toJSON();
```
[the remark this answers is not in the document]
mike-parkhill: yeah, blindly loading everything into memory could be dangerous. Paging it probably makes sense.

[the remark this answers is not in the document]
mike-parkhill: agreed offline to make paging part of the SDK changes rather doing it twice.

src/core/subscan.js
```diff
@@ -20,9 +20,5 @@ export function fetchTransactions({address, page = 0, row = 50}) {
         ...res.data,
         hasNextPage: (page + 1) * row < res.data.count,
       };
-    })
-    .catch(err => {
```
mike-parkhill: we're good to let the error escape?

maycon-mello: Actually I removed this in the latest commit I think

maycon-mello: We should propagate the error

mike-parkhill: yeah, the catch is gone and I was just making sure that allowing the error to bubble out now wasn't going to cause other problems.

mike-parkhill (approved):

PR #97, opened 2021-12-08
fixing issue with sentry not reporting unhandled promises

maycon-mello:


src/core/toast.js
```diff
@@ -27,6 +28,7 @@ export const withErrorToast =
     try {
       await fn(...params);
     } catch (err) {
+      captureException(err);
       showUnexpectedErrorToast();
```
mike-parkhill, line 32: as we capture errors it might be good if we can map common errors to more meaningful messages for the users.

mike-parkhill (approved):

PR #99, opened 2021-12-14
Fixing issues with account export + account backup

maycon-mello:


src/features/accounts/account-slice.js
```diff
@@ -237,6 +237,8 @@ export const accountOperations = {
 
     Logger.debug('Rpc done');
```
mike-parkhill, line 238: feels like this log should be after the sync?

mike-parkhill (dismissed):

src/features/accounts/account-slice.js
maycon-mello: It's waiting for the waitRpcReady to finish. for me it looks correct

maycon-mello: renamed the look to 'waitRpcReady done'

mike-parkhill (approved):

PR #102, opened 2021-12-20
Fix/validate qr code

maycon-mello:


src/features/transactions/transactions-slice.js
```diff
@@ -21,6 +21,16 @@ const initialState = {
   transactions: [],
 };
 
+export const parseTransaction = transaction =>
+  transaction.date instanceof Date
+    ? transaction
+    : {
+        ...transaction,
+        date: new Date(transaction.date),
```
mike-parkhill, line 29: could this blow up if the date is invalid? And if so, is that what we want to happen? Or can we handle it more gracefully?

maycon-mello: In case of an invalid date it will not throw anything. I assume the subscan api will always return a valid date for the transaction. In case we receive an unexpected response from subscan, the react component will throw an exception to sentry, and this error will not propagate to the entire app.

mike-parkhill: can we catch that error a display something meaningful to the user? e.g. "Unable to fetch transaction history. Please try again later" or something

src/core/subscan.js
```diff
@@ -20,5 +23,12 @@ export function fetchTransactions({address, page = 0, row = 50}) {
         ...res.data,
         hasNextPage: (page + 1) * row < res.data.count,
       };
+    })
+    .catch(err => {
+      showToast({
+        message: translate('global.unable_to_fetch_transactions'),
+        type: 'error',
+      });
+      throw err;
```
mike-parkhill, line 32: is the value of continuing to throw just to make sure that we exit from the transaction flow? Or is there something else?

also, are we able to test this function with a mock endpoint?

maycon-mello: Yes, throwing the error will stop the transaction fetch flow. Sure, let me check how to add some unit test on it.

src/core/subscan.test.js
```diff
@@ -0,0 +1,60 @@
+import {setToast} from './toast';
+import {fetchTransactions} from './subscan';
+
+describe('Subscan integration', () => {
```
maycon-mello, line 4: added some unit tests on it

mike-parkhill (approved):

PR #105, opened 2021-12-30
Feat/touch feedback icons

maycon-mello:


src/design-system/buttons.js
```diff
@@ -34,6 +39,11 @@ export const useAsyncCallback = func => {
 export function BackButton(props) {
   const [loading, onPress] = useAsyncCallback(props.onPress || navigateBack);
 
+  // useEffect(() => {
```
mike-parkhill, line 42: commented code - think this is causing the linting errors

mike-parkhill (approved):

PR #115, opened 2022-01-11
Fix/confirm dialog buttons

SamHellawell:


yarn.lock
```diff
@@ -6600,9 +6600,9 @@ combined-stream@^1.0.6, combined-stream@^1.0.8, combined-stream@~1.0.6:
   dependencies:
     delayed-stream "~1.0.0"
 
-"comlinkjs@git://github.com/ronhe/comlink.git#dist":
+"comlinkjs@https://github.com/ronhe/comlink.git#dist":
```
mike-parkhill, line 6603: are these updated intentional?

src/design-system/buttons.js
```diff
@@ -18,16 +20,25 @@ export const runAfterInteractions =
...
 
 export const useAsyncCallback = func => {
+  const isActive = useRef(true);
   const [loading, setLoading] = useState();
   const callback = () => {
     setLoading(true);
     runAfterInteractions(() => {
       Promise.resolve(func()).finally(() => {
-        setLoading(false);
+        if (isActive.currentsActive) {
```
mike-parkhill, line 29: is this ever going to be true? How does it get set?

maycon-mello: The default state is true, once it's unmounted it will be set to false

maycon-mello: but i think it should be `isActive.current` instead

yarn.lock
SamHellawell: yeah, github has disabled this: https://github.blog/2021-09-01-improving-git-protocol-security-github/#no-more-unauthenticated-git

src/design-system/buttons.js
SamHellawell: not sure how this typo happened... fixed

mike-parkhill (approved):

PR #116, opened 2022-01-17
Feat/wallet sdk

maycon-mello:


metro.config.js
```diff
@@ -36,6 +36,7 @@ module.exports = (async () => {
         vm: require.resolve('vm-browserify'),
         src: path.resolve(__dirname, './src'),
         '@docknetwork/react-native-sdk': path.resolve(__dirname, '../dock-rn-web-bridge'),
```
mike-parkhill, line 38: from a naming consistency standpoint should we rename this to `wallet-sdk-react-native` or something?

src/features/accounts/account-slice.js
```diff
@@ -1,7 +1,11 @@
 import {createSlice} from '@reduxjs/toolkit';
-import {WalletRpc} from '@docknetwork/react-native-sdk/src/client/wallet-rpc';
-import {PolkadotUIRpc} from '@docknetwork/react-native-sdk/src/client/polkadot-ui-rpc';
-import {ApiRpc} from '@docknetwork/react-native-sdk/src/client/api-rpc';
+// TODO: Remove usage
```
mike-parkhill, line 2: why do we need to remove this usage? And what's preventing us from doing it now?

src/features/accounts/account-slice.js
```diff
@@ -193,26 +173,9 @@ export const accountOperations = {
...
-          .objects('Account')
-          .filtered('id = $0', account.id)[0];
-
-        if (!cachedAccount) {
-          return;
-        }
-
-        realm.delete(cachedAccount);
-      });
+      Accounts.getInstance().remove(account.id);
```
mike-parkhill, line 178: what's the value of always calling `Accounts.getInstance()` vs maintaining a shared reference to it?

src/features/tokens/price-service.test.js
```diff
@@ -0,0 +1,32 @@
...
+
+    it('expect to handle error', async () => {
+      global.fetch = () =>
+        Promise.resolve({
+          status: 400,
+        });
+
+      const result = await getCoinCapToken('DOCK');
+
+      expect(result).toBe(emptyResponse);
```
mike-parkhill, line 29: should we return more details or throw an exception or something instead of an empty response? An empty response can be ambiguous between "no data found" and "an error occurred". If someone is consuming this SDK and needs to know to show an error to the user it might be easier with an error or a response that includes error details.

mike-parkhill:
looks like yarn is failing in the checks too

src/features/accounts/account-slice.js
maycon-mello: Yeah good point, I've added a reference in the acount-slice. So that we just need to call `Accounts.getInstance()` once.

metro.config.js
maycon-mello: Actually we should remove this line, this is a reference for the old package. Now it was replaced by `@docknetwork/wallet-sdk-core`

src/features/accounts/account-slice.js
maycon-mello: I forgot this todo item, we should avoid using Rpc calls directly in the app, it will be abstracted by the sdk modules.

mike-parkhill:
Still the question about whether to return the empty set vs passing an error back when an error occurs in the `price-service` and the failing check.

src/features/tokens/SendTokenScreen.js
```diff
@@ -298,34 +345,14 @@ export function SendTokenContainer({route}) {
...
-                if (!form.sendMax) {
-                  formUpdates.amountMessage = translate(
-                    'send_token.amount_minus_fees_msg',
-                  );
-                }
-              }
-
-              updateForm({
-                ...formUpdates,
+              console.log('tx fee', fee);
```
mike-parkhill, line 348: are you leaving this log intentionally?

src/rn-rpc-webview/index.js
```diff
@@ -44,7 +44,14 @@ export function RNRpcWebView({onReady}) {
         );
       }}
       onMessage={async event => {
-        const data = JSON.parse(event.nativeEvent.data);
+        let data;
+
+        try {
+          data = JSON.parse(event.nativeEvent.data);
+        } catch (err) {
+          console.log('error to prcess message', event.nativeEvent.data);
```
mike-parkhill, line 52: can you provide a bit more context in the error message - I'm not sure this message would help me get back here in a hurry (other than the `prcess` typo maybe helping in a text search 😸 )

maycon-mello: Actually we should remove it, I was using that just to debug an error but was not helpful.

scripts/link-sdk.js
```diff
@@ -0,0 +1,38 @@
+
+const fs = require('fs');
+const {execSync} = require('child_process')
+const sdkRepoExists = fs.existsSync('../wallet-sdk');
+
+let output;
+
+if (!sdkRepoExists) {
+  console.log('Clonning wallet sdk repo....')
```
mike-parkhill, line 9: typo: `Clonning`

mike-parkhill (commented):
One minor typo and still this question? https://github.com/docknetwork/dock-app/pull/116/files#r786273382

scripts/link-sdk.js
maycon-mello: fixed

maycon-mello:
> 

Should be fixed now

mike-parkhill (approved):

mike-parkhill (approved):

PR #119, opened 2022-03-01
 Close menu when "Apply" is clicked on changing network. 

osinakayah:
Problem: When apply button is clicked, onNetworkChange is called, onNetworkChange is a promise and it takes a while to get resolved, this is the reason the modal does not close immediately.

Solution: Instead of waiting for the promise to get resolved before closing the modal, the modal is closed immediately then the network change process is then triggered next.

Suggestions: Include a toast to notify the user when the network change has been completed (When the promise gets resolved)

Link: https://dock-team.atlassian.net/browse/DCK-2007


src/features/dev-settings/DevSettingsScreen.js
```diff
@@ -119,11 +119,10 @@ export function DevSettingsScreen({onAddAccount, onNetworkChange}) {
...
 
             <Button
-              onPress={async () =>
-                onNetworkChange(networkId).then(() => {
-                  setShowNetworkOptions(false);
-                })
-              }>
+              onPress={async () =>{
+                setShowNetworkOptions(false);
+                onNetworkChange(networkId).then(() => {})
```
mike-parkhill, line 124: what happens if the network change fails? Should we show a toast or something letting them know it failed?

mike-parkhill (changes requested):
Question about how we should handle errors in the network change and it looks like a couple of very minor linting issues.

src/features/dev-settings/DevSettingsScreen.js
osinakayah: Yes, I think we do that too

osinakayah:
@mike-parkhill 
I think we can include a toast message when it is successful and when it fails.
I have pushed another commit for that.

maycon-mello:
Linting issues were fixed in the dev branch and I did a rebase on this one. CI should pass now

src/features/dev-settings/DevSettingsScreen.js
```diff
@@ -119,9 +119,21 @@ export function DevSettingsScreen({onAddAccount, onNetworkChange}) {
...
-                onNetworkChange(networkId).then(() => {})
+                onNetworkChange(networkId)
+                  .then(() => {
+                    showToast({
+                      message: translate('dev_settings.switch_network_success'),
+                      type: 'success',
+                    });
+                  })
+                  .catch(() => {
+                    showToast({
```
mike-parkhill, line 132: would it be useful to log the error? @maycon-mello what scenarios would cause it to fail?

maycon-mello: The substrate server could be offline

maycon-mello: Logging it will send the error to sentry as well, not sure if that's useful

mike-parkhill: Hmm, I don't want to flood Sentry with noise, but if we're seeing consistent issues it would be nice to know why too. If there are known errors we don't need to know about then can we filter them out, but for unknowns log them maybe?

Can we check the quality of the Internet connection if we get an error? Probably wouldn't do it in this ticket, but maybe add it later.

hodlEngineer: @mike-parkhill 
Yes, I think it is possible to check the internet speed.
https://github.com/TSjianjiao/react-native-network-speed
https://github.com/AsbarAli/react-native-network-bandwith-speed

osinakayah: @mike-parkhill 
I think it is possible to check the internet speed.
https://github.com/TSjianjiao/react-native-network-speed
https://github.com/AsbarAli/react-native-network-bandwith-speed

mike-parkhill: and what about filtering out "accepted" errors but allowing others to log?

osinakayah: Maybe we can skip logging errors related to network failures, since most likely they won't be caused by errors in our code but by external factors such as timeout, no network or service not available.

mike-parkhill: Yeah, I guess that makes sense. At some point we may need to manage network connections more intelligently so we can give better messages to the users, but for now we can probably ignore them but should probably track other errors.

osinakayah: @mike-parkhill 
I did a little bit of testing/debugging and no exception is thrown substrate server url is wrong or when there is no network.  It just doesn't do anything.

mike-parkhill: Ok, so does that mean we're not going to change anything here then?

osinakayah: Yes, we do not.

mike-parkhill (approved):

PR #121, opened 2022-03-03
Sort Transaction history in descending order using date

osinakayah:
Problem:  Transaction history is retrieved from redux store and they are not been sorted

Solution: Sort the transactions after getting them from the store.

Link: https://dock-team.atlassian.net/browse/DCK-2005

### DIFF
--- a/src/features/accounts/AccountDetailsScreen.js
+++ b/src/features/accounts/AccountDetailsScreen.js
@@ -178,6 +178,7 @@ function TransactionHistory({accountAddress}) {
       <NBox>
         {transactions.map(item => (
           <TransactionHistoryItem
+            key={item.id}
             transaction={item}
             accountAddress={accountAddress}
           />
--- a/src/features/accounts/AccountDetailsScreen.js
+++ b/src/features/accounts/AccountDetailsScreen.js
@@ -122,6 +122,13 @@ function TransactionHistoryItem({transaction, accountAddress}) {
     </>
   );
 }
+export function sortTransactionHistory(transactions) {
+  return [...transactions].sort((a, b)=>{
+    const aDateTime = a.date ? new Date(a.date).getTime(): new Date().getTime()
+    const bDateTime = b.date ?  new Date(b.date).getTime(): new Date().getTime()
+    return bDateTime-aDateTime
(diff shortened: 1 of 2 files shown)


src/features/accounts/AccountDetailsScreen.test.js
```diff
@@ -30,6 +31,41 @@ describe('AccountDetailsScreen', () => {
     expect(wrapper.dive()).toMatchSnapshot();
   });
 
+  describe('expect transactions to be sorted by date in descending order', () => {
```
mike-parkhill, line 34: what happens if two transactions have the exact same timestamp? Is the result predictable? i.e. what's the secondary sort field?

src/features/accounts/AccountDetailsScreen.js
```diff
@@ -122,6 +122,17 @@ function TransactionHistoryItem({transaction, accountAddress}) {
     </>
   );
 }
+export function sortTransactionHistory(transactions) {
```
mike-parkhill, line 125: This will work fine if all the transactions are already in memory, but do we handle pulling large sets from Subscan using paging or anything? If so, we'll need to get Subscan to handle the sorting.

src/features/accounts/AccountDetailsScreen.test.js
osinakayah: If two or more transactions have exact timestamps, they will be kept side by side.
Currently there is not a secondary sort field, but maybe we can use the amount maybe, keep the transaction with the highest amount at the top maybe.
I am considering if we can use the block height as the secondary sort field if it is available, but there is a chance the block height will also be the same if the timestamps are the same.

src/features/accounts/AccountDetailsScreen.js
osinakayah: Yeah, that will be a more scalable approach. 
Maybe we can have an api service that retrieves all transaction history for an account, the api will handle the sorting and also the filtering of the transaction history.

osinakayah: Yeah, I think that is better approach, alright, I will work on that.

PR #122, opened 2022-03-04
Transaction history not sorted properly

osinakayah:
Problem: Data in realm is not stored in a sorted order.

Solution: Sort the transactions when retrieving them from realm.

Suggestion: Allow other other realm handle other data filtering functions instead of handling them in JS.
Link: https://dock-team.atlassian.net/browse/DCK-2005


mike-parkhill (commented):
can we unit test this?

osinakayah:
> can we unit test this?

Yeah, no problem.e

osinakayah:
> 685a738



> can we unit test this?

Unit test has been included

src/features/accounts/AccountDetailsScreen.test.js
```diff
@@ -5,8 +5,66 @@ import {
...
+    migration: () => {
+      // No migration required so far
+    },
+  });
+};
+
+const initMockTransactions = () => {
+  return [
+    {
+      amount: '10',
```
mike-parkhill, line 34: is there value in including a record that should be filtered in the data?

src/features/transactions/transactions-slice.js
```diff
@@ -39,9 +39,9 @@ const transactions = createSlice({
       state.loading = action.payload;
     },
     setTransactions(state, action) {
-      state.transactions = action.payload
-        .map(parseTransaction)
-        .sort(sortTransactions);
+      console.log(action);
```
mike-parkhill, line 42: do we still need this log?

src/features/transactions/transactions-slice.js
```diff
@@ -104,38 +104,50 @@ export const transactionsOperations = {
...
     try {
       do {
-        data = await fetchTransactions({address: account, page});
-        data.transfers.forEach(handleTransaction);
-        page++;
+        data = await fetchTransactions({
+          address: account,
+          page,
+        });
+        if (Array.isArray(data.transfers)) {
```
mike-parkhill, line 111: are we guaranteed that if `data.transfers` is NOT an array that `data.hasNextPage` will be false? Just wondering if we could ever get stuck in a loop if bad data is returned.

src/features/transactions/transactions-slice.js
```diff
@@ -104,38 +104,50 @@ export const transactionsOperations = {
...
+        const realmTransactions = realm
+          .objects('Transaction')
+          .filtered("!(status == 'complete' && hash != null)")
+          .sorted('date', shouldReverseOrder)
+          .toJSON();
+        dispatch(transactionsActions.setTransactions(realmTransactions));
+      } else {
+        const realmTransactions = realm
+          .objects('Transaction')
+          .sorted('date', shouldReverseOrder)
```
mike-parkhill, line 146: why don't we filter in Testnet?

osinakayah: I think when on mainnet and we fetch the transactions from subscan, it is possible some transactions does  not returning any transaction hash. Probably because the transaction is yet to be confirmed, So I think that is the reason we were filtering out such transactions.

But on testnet that scenario is unlikely to happen

osinakayah: I think when on mainnet and we fetch the transactions from subscan, it is possible some transactions does  not returning any transaction hash. Probably because the transaction is yet to be confirmed, So I think that is the reason we were filtering out such transactions.

But on testnet that scenario is unlikely to happen

src/features/accounts/AccountDetailsScreen.test.js
osinakayah: The mock transactions records is focused on testing the sorting logic.

src/features/transactions/transactions-slice.js
osinakayah: No, not at all.

src/features/accounts/AccountDetailsScreen.test.js
mike-parkhill: do we have a test that covers filtering?

src/features/transactions/transactions-slice.js
mike-parkhill: Maybe so, but I'm just wondering if it's worth the slightly extra complexity of having different paths in the code for testnet and mainnet. Even if the filter never does anything in testnet I don't see a down side to running the same filters in both places. I doubt it's much (if any) overhead.

osinakayah: @mike-parkhill 
Apart from the filter, when on mainnet we need to make an API request to fetch transactions from subscan, so we need the different paths.

mike-parkhill: you could pull the realm part out of the mainnet block though since it's common to both 

```
if (mainnet)
{
... api stuff ...
}

... common realm stuff...
```

osinakayah: @mike-parkhill 
Yeah, we can do that. I just updated the PR

mike-parkhill (approved):

PR #128, opened 2022-03-09
increment build number

maycon-mello:


mike-parkhill (approved):
really wish we could do this automagically

maycon-mello:
> really wish we could do this automagically

There are some cases we need to update the app version. For instance, v0.1.6 to v.0.1.7, it will be done for each production release.

But for incremental build like this one, which will just increment the build number its easy to add that on the CI

PR #133, opened 2022-03-21
Chore/v2 navigation

maycon-mello:


src/design-system/index.js
```diff
@@ -57,11 +64,77 @@ const styles = StyleSheet.create({
...
 
-export function ScreenContainer({children, hideGlobalHeader, ...props}) {
+const menuOptions = [
+  {
+    id: 'tokens',
+    route: Routes.ACCOUNTS,
+    name: translate('app_navigation.tokens'),
+    Icon: MenuTokensIcon,
+  },
+  {
```
mike-parkhill, line 75: if you hide this member behind a feature flag is that the only spot you'd need to update?

mike-parkhill: I'm fine with the new menu being available early for the existing functionality.

mike-parkhill (commented):
Waiting for the feature flag, but otherwise I think is fine.

src/features/accounts/AccountsScreen.js
```diff
@@ -54,7 +54,7 @@ export const AccountsScreen = withErrorBoundary(
     const [showImportAccount, setShowImportAccount] = useState();
 
     return (
-      <ScreenContainer testID={AccountsScreenTestIDs.screen}>
+      <ScreenContainer testID={AccountsScreenTestIDs.screen} showTabNavigation>
```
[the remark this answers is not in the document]
maycon-mello: I think the feature flag should be handled in the tab navigation only. There are some screens which the showTabNavigation will not be displayed (unlock, create wallet...), so it's important to have it here.

src/features/dev-settings/DevSettingsScreen.js
```diff
@@ -27,36 +27,36 @@ import {
...
-    },
-  );
 
   const [networkId, setNetworkId] = useState(currentNetworkId);
 
   useEffect(() => {
     setNetworkId(currentNetworkId);
   }, [currentNetworkId]);
 
+  console.log('features', features);
```
mike-parkhill, line 56: debug log?

src/features/dev-settings/DevSettingsScreen.js
```diff
@@ -100,13 +100,22 @@ export function DevSettingsScreen({
...
-        onPress: () => {
-          setShowTransactionHistory(true);
-        },
+        value: features.showTestnetTransaction,
+        isSwitch: true,
+        onPress: () => onFeatureToggled(Features.showTestnetTransaction),
       });
     }
+
+    options.push({
```
mike-parkhill, line 107: if your features object included a translation key you could probably iterate over the array rather than hard-coding the features here

it would be awesome if you could simply update the `feature-flags.js` and then it magically showed up here

src/features/dev-settings/DevSettingsScreen.js
```diff
@@ -181,45 +190,30 @@ export function DevSettingsScreen({
...
-                  label={translate('dev_settings.show_testnet_transaction')}
-                  value={true}
-                />
-                <Select.Item
-                  label={translate('dev_settings.hide_testnet_transaction')}
-                  value={false}
-                />
-              </Select>
-            </Stack>
+        {/*{showTransactionHistory ? (*/}
```
mike-parkhill: commented code

src/features/dev-settings/DevSettingsScreen.js
```diff
@@ -286,29 +280,27 @@ export function DevSettingsScreen({
...
-  const handleToggleTransactionHistory = useCallback(
-    testnetTransactionConfig => {
-      return dispatch(
-        appActions.setShowTestnetTransaction(testnetTransactionConfig),
-      );
-    },
-    [dispatch],
-  );
+  const handleFeatureToggled = (name: String) => {
+    console.log('toggle features', name);
```
mike-parkhill, line 262: debug log?

mike-parkhill (commented):
for the debug logs, if you want to keep them I'm okay with that , just calling them out - and possibly we should use the proper logging framework there, right?

src/features/dev-settings/DevSettingsScreen.js
maycon-mello: That was a great suggestion. I've pushed an update in the PR, it's so easy to add new features now. Thanks!

maycon-mello:
> for the debug logs, if you want to keep them I'm okay with that , just calling them out - and possibly we should use the proper logging framework there, right?

Yeah, I just forgot those debug logs and i've removed that already. But will try to start using the Logger.debug instead next time.

mike-parkhill (approved):

PR #135, opened 2022-03-22
Made the share and copy button visible on the receive dock screen

osinakayah:
https://dock-team.atlassian.net/browse/DCK-1823


maycon-mello (dismissed):

src/features/tokens/__snapshots__/ReceiveTokenScreen.test.js.snap
```diff
@@ -49,7 +49,7 @@ exports[`ReceiveTokensScreen should render correctly 1`] = `
         p={2}
       >
         <QRCode
-          size={525}
+          size={375}
```
mike-parkhill, line 52: if we're shrinking the size of the QR code are we going to run into any problems fitting the data we need into it?

PR #140, opened 2022-03-25
Modify error message

osinakayah:
https://dock-team.atlassian.net/browse/DCK-2028

### DIFF
--- a/src/features/wallet/wallet-slice.test.js
+++ b/src/features/wallet/wallet-slice.test.js
@@ -1,6 +1,5 @@
 import {translate} from 'src/locales';
-import {validateAndImport} from './wallet-slice';
-import {WalletRpc} from '@docknetwork/react-native-sdk/src/client/wallet-rpc';
+import {validateWalletImport} from './wallet-slice';
 
 describe('Wallet Slice', () => {
   it('expect error message to be accurate', () => {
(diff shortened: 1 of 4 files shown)


osinakayah:
When trying to import a wallet from a JSON file, the "wallet.importWallet" method throws an exception if the JSONdata or the password provided is incorrect.

But it does not say what specifically went wrong, if it was the password or the encryptedJSON data.


So I made the error message more generic too.

It is kind of similar to when a user is trying to authenticate with username or password, if one of the details is incorrect, it does not tell you which, but just says one of them is incorrect.

osinakayah:
> Do we have any tests around this function that need to be updated (and/or can we add one while we're in here)?

@mike-parkhill 
Currently, we do not have test for translations, but we can include that.

src/features/wallet/wallet-slice.test.js
```diff
@@ -0,0 +1,11 @@
+import {translate} from 'src/locales';
+
+describe('Wallet Slice', () => {
+  it('expect error message to be accurate', () => {
```
mike-parkhill, line 4: I was thinking we'd test the error condition rather than the translation. So call `validateAndImport()` with various inputs to make sure we get the expected outputs for both success and error conditions.

To make testing easier it might make sense to separate the validate logic from the import logic into separate functions.

src/features/wallet/wallet-slice.test.js
```diff
@@ -8,4 +9,20 @@ describe('Wallet Slice', () => {
...
+
+    return validateWalletImport(fileData, 'correct');
+  });
+
+  it('expect to validate import when invalid', () => {
```
mike-parkhill, line 20: we have at least two conditions that are invalid -> docs == 0 or accounts ==0, right? Can we test for both scenarios?

osinakayah: Got it.

mike-parkhill (approved):

PR #141, opened 2022-03-29
Feat/credential management

maycon-mello:


src/features/credentials/CredentialsScreen.js
```diff
@@ -5,14 +5,78 @@ import {
...
+        fontFamily={Theme.fontFamily.nunitoSans}>
+        {translate('credentials.empty_items')}
+      </Text>
+    </Center>
+  );
+}
+
+function CredentialListItem({credential}) {
+  return (
+    <NBox bgColor="#000000" p={4} borderRadius={10} m={2}>
```
maycon-mello, line 41: Working to add theme variables

src/features/credentials/test-credential.json
```diff
@@ -0,0 +1,42 @@
+{
```
maycon-mello, line 1: I think we can move this to the wallet-sdk repo

src/assets/icons/empty-credential.svg
```diff
@@ -0,0 +1,7 @@
+<svg width="41" height="40" viewBox="0 0 41 40" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
maycon-mello: I can check that, for now I'm just following the figma designs

src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -0,0 +1,103 @@
...
+
+    it('expect to trigger handler1', async () => {
+      const data = 'handler1';
+      await qrCodeHandler(data, handlers);
+
+      expect(handler1).toBeCalledWith(data);
+      expect(handler2).not.toBeCalled();
+    });
+
+    it('expect to trigger handler1', async () => {
```
mike-parkhill, line 25: duplicated test name

src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -0,0 +1,103 @@
...
+  });
+  describe('addressHandler', () => {
+    it('expect to ignore invalid data', async () => {
+      jest.spyOn(UtilCryptoRpc, 'isAddressValid').mockReturnValueOnce(false);
+
+      navigationRef.current = {
+        navigate: jest.fn(),
+      };
+
+      await addressHandler('some-address');
```
mike-parkhill, line 41: should probably assert that it returns true/false properly too

src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -0,0 +1,103 @@
...
+
+    it('expect to navigate to send tokens route', async () => {
+      jest.spyOn(UtilCryptoRpc, 'isAddressValid').mockReturnValueOnce(true);
+
+      navigationRef.current = {
+        navigate: jest.fn(),
+      };
+
+      const address = 'some-address';
+      await addressHandler(address);
```
mike-parkhill, line 54: same here

src/features/qr-code-scanner/qr-code.js
```diff
@@ -20,8 +20,17 @@ export async function credentialHandler(data) {
...
+      credentialData = await credentials.getCredentialFromUrl(data);
+      await credentials.add(credentialData);
+    } else {
+      const jsonData = JSON.parse(data);
+      credentialData = await credentials.add(jsonData);
+    }
+
     navigate(Routes.APP_CREDENTIALS);
     return true;
   } catch (err) {
```
mike-parkhill, line 36: we should do something more with this error handler.  Right now we'd have no idea what went wrong or even if it was attempting to resolve a URL or JSON.

src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -91,9 +94,31 @@ describe('qr-code', () => {
...
+      const result = await credentialHandler('http://some-url');
+
+      expect(result).toBeTruthy();
+      expect(navigationRef.current.navigate).toBeCalledWith(
+        Routes.APP_CREDENTIALS,
+        undefined,
+      );
+    });
+
+    it('expect to handle json data', async () => {
```
mike-parkhill, line 106: can we add a test with malformed JSON?

mike-parkhill: and maybe one with valid JSON that's not a credential

src/features/qr-code-scanner/qr-code.js
```diff
@@ -34,6 +63,13 @@ export async function credentialHandler(data) {
...
+    if (isUrl) {
+      console.error(`Unable to resolve url: ${data}`);
+    } else {
+      console.error(`Unable to resolve json: ${JSON.stringify(data)}`);
```
mike-parkhill, line 69: if there's an error above could `JSON.stringify(data)` blow up?

maycon-mello: Actually I don't see how the JSON.stringify could  blow up. Even for undefined value it will not break

mike-parkhill: guess I was worried if the parse failed then stringify might too - e.g. circular refs or something

maycon-mello: Ah good point, circular refs can break. Let me create a utility function for that

src/features/credentials/CredentialsScreen.js
mike-parkhill: this is done, right?

src/core/index.js
```diff
@@ -0,0 +1,13 @@
...
+ * JSON.stringify with error handler
+ *
+ * @param value
+ * @returns {string}
+ */
+export function safeJSONStringify(value) {
+  try {
+    return JSON.stringify(value);
+  } catch (err) {
+    return err.toString();
```
mike-parkhill, line 11: this could be confusing/misleading to someone calling this method. The caller might see a string got returned and assume it's all good and use the string as if it's valid JSON when it's actually an error message.

maybe rename to `getJsonOrError` or something..

or use an approach like this one: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Cyclic_object_value

src/features/credentials/CredentialsScreen.js
maycon-mello: Yes

mike-parkhill (approved):

PR #143, opened 2022-04-04
Fix/qr scanner

maycon-mello:


src/features/qr-code-scanner/qr-code.js
```diff
@@ -20,8 +20,17 @@ export async function credentialHandler(data) {
   try {
     // @ts-ignore
     const credentials: Credentials = Credentials.getInstance();
-    const credentialData = credentials.getCredentialFromUrl(data);
-    await credentials.add(credentialData);
+
```
mike-parkhill, line 23: could probably add a null/string check up here and skip the rest if it doesn't pass that simple check.

maycon-mello:
This has https://github.com/docknetwork/dock-app/pull/141 as a dependency

src/features/qr-code-scanner/qr-code.js
maycon-mello: Not sure on this, are you referring to the `Credentials.getInstance()`?

maycon-mello: Ah got it, it's the handler. I've pushed an updated for that

maycon-mello:
Added a new commit @mike-parkhill, with better error handlers and tests

src/features/credentials/credentials.js
```diff
@@ -18,6 +18,7 @@ export function useCredentials() {
     await syncCredentials();
   };
 
+  console.log('credetials', items);
```
mike-parkhill, line 21: probably meant to delete this?

src/features/qr-code-scanner/qr-code.js
```diff
@@ -34,16 +63,40 @@ export async function credentialHandler(data) {
...
-      return;
+      return true;
     }
   }
+
+  return false;
+}
+
+export async function qrCodeHandler(data, handlers = qrCodeHandlers) {
+  const success = await executeHandlers(data, handlers);
```
mike-parkhill, line 94: validate `data` is defined?

maycon-mello: its validated inside of executeHandlers already

src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -30,6 +42,27 @@ describe('qr-code', () => {
...
+
+    it('expect not to trigger handlers for empty data', async () => {
+      const data = '';
+      const mockHandlers = [jest.fn()];
+      await qrCodeHandler(data, mockHandlers);
+
+      expect(mockHandlers[0]).not.toBeCalled();
+    });
+
+    it('expect not to show error toast for not handled data', async () => {
```
mike-parkhill, line 54: do we give the user any indication that something went wrong?

mike-parkhill (approved):

PR #144, opened 2022-04-04
Fix/document picker issue

maycon-mello:


src/core/storage-utils.js
```diff
@@ -0,0 +1,13 @@
+import DocumentPicker from 'react-native-document-picker';
+
+export function pickDocuments() {
+  return DocumentPicker.pick({
+    type: [DocumentPicker.types.allFiles],
+  }).catch(err => {
+    if (err.code === 'DOCUMENT_PICKER_CANCELED') {
```
mike-parkhill, line 7: do you ever need to be able to differentiate `cancel` vs `no docs found`?

maycon-mello: No, `no docs found` will require the user to cancel the document picker

mike-parkhill (approved):

PR #155, opened 2022-04-15
Included google analytics

osinakayah:
https://dock-team.atlassian.net/browse/DCK-2133


mike-parkhill (commented):
should we be injecting our keys rather than hard-coding them?

what data does this publish? Have you been able to test it?

osinakayah:
> should we be injecting our keys rather than hard-coding them?
>  The google-services.json file and GoogleService don't contain sensitive information, but on second thought though, we can add them in .gitignore
> what data does this publish? Have you been able to test it?


It publishes data about the number of users we have, their demographic, age, country, where they are accessing it from, the version they have, the phone model etc.

mike-parkhill:
Cool, yep I see it in the new project I created on GA. Looks a bit limited in details, but we can no doubt flesh that out in the future.

mike-parkhill:
what about this question?

> should we be injecting our keys rather than hard-coding them?

mike-parkhill (approved):
Thanks for moving the credentials out.

PR #156, opened 2022-04-18
add import from json

maycon-mello:


src/core/storage-utils.js
```diff
@@ -11,3 +14,27 @@ export function pickDocuments() {
...
 }
+
+export async function pickFileData() {
+  const files = await pickDocuments();
+
+  if (!files.length) {
+    return;
+  }
+
+  return RNFS.readFile(files[0].fileCopyUri);
```
mike-parkhill, line 25: should we validate that `fileCopyUri` is actually populated? Or are we safe to assume it always will be?

mike-parkhill: related, if `readFile` fails, does that blow things up?

src/features/credentials/credentials.js
```diff
@@ -18,9 +21,23 @@ export function useCredentials() {
...
+      await Credentials.getInstance().add(jsonData);
+    } catch (err) {
+      showToast({
+        message: translate('credentials.invalid_credential'),
+        type: 'error',
+      });
+    }
+  };
+
   console.log('credetials', items);
```
mike-parkhill, line 37: console.log

src/core/storage-utils.js
maycon-mello: fileCopyUri could break in case we upgrade some dependency, or the device gets updated to some new version. Let me add proper validation, so that we can monitor it from sentry

maycon-mello: readFile can blow things up in case the app doesn't have proper permissions. Let me do the same, adding sentry alerts for it, so that we can monitor cases of bad permissions

maycon-mello: Maybe a helper function to wrap `readFile` and `writeFil`, which will be used across the app

src/core/storage-utils.js
```diff
@@ -15,19 +16,38 @@ export function pickDocuments() {
   });
 }
 
+export function readFile(path) {
+  assert(!!path, 'file path is required');
+
+  try {
+    return RNFS.readFile(path);
+  } catch (err) {
+    throw new Error(`Unable to read file ${path}`);
```
mike-parkhill, line 25: should we embed the original error so we don't lose why it failed?

maycon-mello: maybe log it

mike-parkhill (approved):

PR #159, opened 2022-04-19
Test deployment

osinakayah:


mike-parkhill (commented):
Looks good to me, but I'm no expert on fastlane. I'd like @maycon-mello to look too.

maycon-mello (approved):

maycon-mello:
Looks great, we are almost there on getting automatic deployments 🤩

PR #163, opened 2022-04-26
Test credential has required properties

osinakayah:


src/features/wallet/wallet-slice.js
```diff
@@ -55,11 +55,32 @@ export const walletSelectors = {
   getCreationFlags: state => getRoot(state).creationFlags || {},
 };
 
+export const validateWalletKeys = walletJSONData => {
```
mike-parkhill, line 58: The blockchain sdk doesn't have a validation function we could borrow does it? Sam or Lovesh might know.

Assuming there isn't already an existing validation function should this one verify the values as well as the keys? i.e. make sure the keys have defined values that are valid

osinakayah: >The blockchain sdk doesn't have a validation function we could borrow does it? Sam or Lovesh might know.
I asked @maycon-mello and he said we did not have a validation function, that we will have to include it. I will ask Sam and Lovesh also.

>Assuming there isn't already an existing validation function should this one verify the values as well as the keys? i.e. make sure the keys have defined values that are valid

Based on my research the keys I listed that are required keys, and the values the contained in some of them is kind of flexible. Except for the "issuer", that I think should be a valid existing DID, and "issuanceDate" that has be  in the past.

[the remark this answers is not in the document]
osinakayah: Yeah. That make sense. Let me try and come up with a better descriptive name.

osinakayah: I confirmed from Sam and Lovesh, the blockchain cannot verify the credential if it has not been signed.

src/features/wallet/wallet-slice.test.js
```diff
@@ -75,4 +79,60 @@ describe('Wallet Slice', () => {
...
+          tag: 'Yk9KFp7RZWa3m2sYtUnRPQ',
+        },
+      },
+    };
+    expect(verifyJSONCredentialFormat(validWalletData)).toBeTruthy();
+  });
+
+  it('expect false invalid wallet properties', () => {
```
mike-parkhill, line 124: theoretically we should probably have failure examples for each required field

src/features/wallet/wallet-slice.js
SamHellawell: cannot run full validation but we could detect for required properties to exist at minimum (credentialsubject, context etc)

PR #166, opened 2022-04-27
load credentials after import from json

maycon-mello:


mike-parkhill (commented):
can we test this to make sure it always happens?

mike-parkhill (approved):

PR #171, opened 2022-04-29
Fix/add fallback logo to credentials

maycon-mello:


src/features/credentials/CredentialsScreen.js
```diff
@@ -126,15 +127,22 @@ const CredentialListItem = withErrorBoundary(({credential, onRemove}) => {
...
+              width={50}
+              height={50}
+              alt={''}
+              source={{
+                uri: credential.issuer.logo,
+              }}
+            />
+          ) : (
+            <PolkadotIcon
+              address={getDIDAddress(credential.issuer.id)}
```
mike-parkhill, line 142: do we need to worry about an empty/undefined `id`?

src/features/credentials/credentials.js
```diff
@@ -59,6 +59,12 @@ Credentials.getInstance().wallet = {
   remove: params => WalletRpc.remove(params),
 };
 
+export function getDIDAddress(did) {
+  assert(!!did, 'did is required');
```
mike-parkhill, line 63: ah, you're checking it here - nvm.

src/features/credentials/credentials.test.js
```diff
@@ -64,6 +65,23 @@ describe('Credentials helpers', () => {
     });
   });
 
+  it('getDIDAddress', () => {
```
mike-parkhill, line 68: should test the assertion too by passing in null

maycon-mello: There is a test for it in the line 69

mike-parkhill: damn, so there is. 🤦

mike-parkhill (approved):

PR #172, opened 2022-05-02
Custom analytics event

osinakayah:


src/features/analytics/analytics-slice.js
```diff
@@ -0,0 +1,37 @@
...
+      .logEvent(eventName, params)
+      .catch(err => {
+        captureException(err);
+      });
+  }
+}
+export const ANALYTICS_EVENT = {
+  TOKENS: {
+    SEND_TOKEN: 'token_send',
+    RECEIVE_TOKEN: 'token_receive',
```
mike-parkhill, line 16: I don't see this being used

src/features/analytics/analytics-slice.js
```diff
@@ -0,0 +1,37 @@
+import analytics from '@react-native-firebase/analytics';
+import {captureException} from '@sentry/react-native';
+
+export function logAnalyticsEvent(eventName, params) {
```
mike-parkhill, line 4: should this be async? I'm thinking most the time we should be able to fire and forget - telemetry succeeding or not shouldn't affect the behaviour of the app from a user perspective

osinakayah: Yes, this way it won't block the user action in any sort of way.

src/features/analytics/analytics-slice.js
osinakayah: I thought about it, the "token_receive" event is not user initiated action.
The best we can do is, send the "token_receive" event when the user clicks on the "Receive" button in the app to get his wallet address, but it does not necessarily mean the user has receive any token.

mike-parkhill: From user activity standpoint though it might be useful to know if they're trying to receive tokens in the app. Telemetry is as much about what the user tried to do as it is about what they succeeded in doing.

osinakayah: Got it. I will add analytics when the click on the "Copy" button or the "Share button" in the receive screen.

mike-parkhill (dismissed):

PR #179, opened 2022-05-02
Select transak settings based on network

osinakayah:


src/features/trade/components/TransakProvider.js
```diff
@@ -73,15 +78,49 @@ export default function TransakPaymentProvider({
 }) {
   const [buyState, setBuyState] = useState(BUY_STATES.INTRO);
 
+  const currentNetworkId = useSelector(appSelectors.getNetworkId);
+
+  const TRANSAK_ENVIRONMENT = useMemo(
```
mike-parkhill, line 83: this works, but I think I would have gone with an env object that bundled everything for one env together and then a single array. Probably just a preference thing though.

`
envs = {
    mainnet: {
       env: PROD_TRANSAK_ENVIRONMENT,
       api_key: PROD_TRANSAK_API_KEY,
       base_url: PROD_TRANSAK_BASE_URL
    }, ....
}
`

and then lookups would be `envs[currentNetworkId].base_url`, etc.

mike-parkhill (commented):
Is there a unit test or something you could write to verify that the right values are being selected based on the network?

mike-parkhill (approved):

PR #180, opened 2022-05-03
Update wallet app to match new theming of the Dock website

osinakayah:


src/design-system/theme.js
```diff
@@ -6,6 +6,7 @@ export const Theme = {
   fontFamily: {
     nunitoSans: 'Nunito Sans',
     montserrat: 'Montserrat',
+    satoshi: 'Satoshi',
```
mike-parkhill, line 9: why not have a `Theme.fontFamily.default` that points at `satoshi`? If we used that everywhere if/when we switch font again it means updating one spot instead of fifty.

src/design-system/theme.js
```diff
@@ -33,12 +34,18 @@ export const Theme = {
     transactionPending: '#FBBF24',
     qrCodeBackground: '#fff',
     credentialCardBg: '#000',
+    absoluteBlack: '#000',
```
mike-parkhill, line 37: previously we were mostly labeling colours based on their usage (e.g. primaryBackground), why switch to colour specific naming now?

src/design-system/theme.js
osinakayah: That actually make sense. No problem, I will do quick refactor.

src/design-system/theme.js
osinakayah: I felt this way is a bit more generic and prevent color duplicates. Because text, buttons and screen background can be absolute black.

If I had done "primaryBackground" => "#000",
and "activeTextColor" => "#000".

Using this method, the color codes get duplicated.

mike-parkhill: it means you have more places to update though in the future if we decide to change all buttons to green instead of absoluteBlack. Makes it easier to miss one.

My understanding of theming is that you use usage names even though you might end up with the same colour for more than one because it's easier to change your whole theme later by just updating the theme file and not having to touch the actual code

mike-parkhill (approved):

PR #182, opened 2022-05-03
Sent and receive tab in the transaction history is showing records two times

osinakayah:
https://dock-team.atlassian.net/browse/DCK-2203


mike-parkhill (commented):
any tests you can write to prove it?

mike-parkhill (approved):

PR #185, opened 2022-05-05
Remove back button

osinakayah:


src/features/credentials/CredentialsScreen.js
```diff
@@ -196,6 +185,7 @@ export function CredentialsScreen({credentials, onRemove, onAdd}) {
 
 export function CredentialsContainer(props) {
   const {credentials, handleRemove, onAdd} = useCredentials();
+  console.log(credentials, 'credentials');
```
maycon-mello, line 188: Can we remove this log?

PR #188, opened 2022-05-06
Chore/enable v2 flags

maycon-mello:


ios/DockApp.xcodeproj/project.pbxproj
```diff
@@ -787,15 +787,15 @@
 				CODE_SIGN_ENTITLEMENTS = DockApp/DockApp.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
 				CODE_SIGN_STYLE = Manual;
-				CURRENT_PROJECT_VERSION = 131;
```
mike-parkhill, line 790: version numbers meant to be different? 136 vs 137?

mike-parkhill (approved):

ios/DockApp.xcodeproj/project.pbxproj
maycon-mello: Good catch, it should be the same. Would create an issue in the next deployment

PR #190, opened 2022-05-09
Show scanner only when screen is focused

osinakayah:


src/features/qr-code-scanner/QRScanScreen.js
```diff
@@ -82,22 +83,22 @@ const QRCodeContainer = styled.View`
...
-          style={styles.scanner}
-          reactivate={true}
-          reactivateTimeout={10000}
-          flashMode={RNCamera.Constants.FlashMode.off}
-          cameraStyle={{height: Dimensions.get('window').height}}
-          topViewStyle={styles.scannerTopView}
-          bottomViewStyle={styles.scannerBottomView}
-          onRead={event => onData(event.data)}
-        />
+        {isScreenFocus && (
```
maycon-mello, line 90: This could potentially break in case isScreenFocus is a string or number. Ideally we should wrap isScreenFocus with `Boolean(isScreenFocus)`

maycon-mello (changes requested):
added a comment

src/features/qr-code-scanner/QRScanScreen.js
osinakayah: @maycon-mello 
useIsFocused hook returns a boolean from the documentation.

https://reactnavigation.org/docs/use-is-focused/

mike-parkhill: since this function is exported some other code could call it where you can't guarantee that `useIsFocused` was the source of the value. Adding a cheap check will protect you.

osinakayah: That make sense. Got it.

maycon-mello (approved):

PR #227, opened 2022-05-13
QR auth link scanning + debug credential

SamHellawell:


src/features/qr-code-scanner/qr-code.js
```diff
@@ -61,7 +62,60 @@ export async function credentialHandler(data) {
...
+        message: `Sign in error: ${e.message}`,
+      });
+    }
+    return true;
+  }
+
+  return false;
+}
+
+export const qrCodeHandlers = [authHandler, addressHandler, credentialHandler];
```
mike-parkhill, line 120: can we put the `authHandler` behind a feature flag?

SamHellawell: i dont think we need to as no users will be exposed to the auth system

maycon-mello (approved):

src/features/qr-code-scanner/qr-code.js
SamHellawell: its not like there is any kind of UI to interact with it

PR #233, opened 2022-05-18
fixing issue with credential issuance date format

maycon-mello:


src/features/credentials/credentials.test.js
```diff
@@ -82,6 +83,21 @@ describe('Credentials helpers', () => {
     ).toBe('5CNyqnHYmrbSE9nmQnpyhdZHi1TavExi3kFWbfrRh1WxQw6A');
   });
 
+  describe('processCredential', () => {
+    it('expect to remove timezone offset', () => {
+      const issuanceDate = new Date('2022-04-24T00:53:13.265Z');
+      const credential = {issuanceDate};
+      const result = processCredential({content: credential});
+      expect(result.content.issuanceDate.getDate()).toBe(24);
+      expect(result.content.issuanceDate.getMonth()).toBe(3);
```
mike-parkhill, line 92: removing the timezone offset changes the month when we're on the 24th? This doesn't feel right.

mike-parkhill: ah, ok - that's not confusing at all 😛

mike-parkhill (approved):

PR #235, opened 2022-05-23
Format date

osinakayah:


src/features/credentials/CredentialsScreen.js
```diff
@@ -39,7 +39,11 @@ function renderObjectAttributes(credential) {
             {Object.keys(attr).map(attrKey => {
               const item = attr[attrKey];
 
-              return <Text fontSize={'12px'}>{item}</Text>;
+              return (
+                <Text fontSize={'12px'}>
+                  {isValidDate(item) ? formatDate(item) : item}
```
mike-parkhill, line 44: does displaying random text/numbers in the date field make sense? 

of course, if we don't, we'll have to somehow communicate to the user that there's an invalid value... leaving it might be simpler? 

Thoughts?

osinakayah: Yeah, I actually thought about that, ideally it might not be a good idea to format any information inside of the credential attributes, displaying it how it was inputted seems to make more sense.

maycon-mello (approved):

src/features/credentials/CredentialsScreen.js
mike-parkhill: but wasn't the point of this ticket to do that formatting?

My question was more around the ternary operator's "else" block where we display the `item` as-is.

maycon-mello (changes requested):
agreed with Mike, seems like the formatting is missing

src/features/credentials/CredentialsScreen.js
osinakayah: The "item" variable can be anything. It can be a date, it can be a normal string value, it can be a number, it is not must for it to be a date.
So if it is a date, I try to format it as a date, but if it is not a date, I display the variable contained inside as it is.

mike-parkhill: your latest change got rid of the formatDate() call though...

osinakayah: @mike-parkhill 
@maycon-mello 
My reason for removing it is, the attributes are just key pair value that was included in the credential, it might be a good idea to display it the way it was added. For example, we might have a value like this 
{
referenceId: "2011-04-04"
},

Dock wallet will end up formatting it as a date, because the string format is similar to a date. But it does not necessarily mean the person who created has intentions for it to be a date.

osinakayah: "credentialSubject": {
        "degree": {
            "name": "Röbber Schmidthü",
            "type": "Bs in cs",
            "dateEarned": "2022-05-03"
        },
        "person": {
            "fullName": "SchmidtÜ",
            "dateOfBirth": "1984-04-26"
        },
        "id": "Markus"
    }
    
    This what it looks like, the data in credentialSubject, I think it can be anything, as long as it is a valid map.

SamHellawell:
i would advise that we close this PR as the prettyvc credential display update will cover this

PR #237, opened 2022-05-31
Create a single Default DID

osinakayah:


src/features/didManagement/didManagment-slice.js
```diff
@@ -0,0 +1,59 @@
...
+  await wallet.add({
+    ...didDocumentResolution,
+  });
+  await wallet.update({
+    ...keyDoc,
+    correlation: correlations,
+  });
+};
+
+const initializeDiDs = async () => {
```
mike-parkhill, line 32: nitpick: inconsistent use of DID vs DiD

src/features/didManagement/didManagment-slice.js
```diff
@@ -0,0 +1,59 @@
...
+  if (keyDocs.length > 0) {
+    const didResolutionDocuments = await wallet.query({
+      type: 'DIDResolutionResponse',
+    });
+
+    if (didResolutionDocuments.length === 0) {
+      await createDID(keyDocs[0]);
+    }
+  } else {
+    const keyDoc = await createKeyDoc();
```
mike-parkhill, line 46: Think I'm not understanding what's happening here. Are we creating a new keyDoc every time we load the wallet after the first time? Why? And why don't we create one the first time?

osinakayah: No, we only create it if there is non.
Each time the wallet is loaded, we check if we have a keyDoc, if we don't we, create one, then we check if there is didDocument, if there is non, we create a didDocument.

mike-parkhill: right, I see that now.

Thanks.

mike-parkhill (commented):
Just linting issues left I think.

PR #240, opened 2022-06-03
Update app version from 0.2 to 0.2.1

maycon-mello:


maycon-mello:
> I thought we had this scripted?

We have the build number scripted. Updating app versions is something manual, and will happen after each production release.

maycon-mello:
> 

It would be hard to script that, it depends on the type of features we will release. In case of major features we might need to update from 0.2 to 0.3, for hot fixes we might go from 0.2 to 0.2.1. So would be hard to script this version updates

PR #241, opened 2022-06-06
Dcka 758 deep linking

osinakayah:


mike-parkhill (commented):
are all the plist and manifest file changes intentional?

PR #246, opened 2022-06-07
fixing issue to import accounts

maycon-mello:


src/features/account-creation/create-account-slice.js
```diff
@@ -150,7 +150,7 @@ export const createAccountOperations = {
       }
 
       const address = await keyringService.addressFromUri({
-        phrase,
+        mnemonic: phrase,
```
maycon-mello, line 153: parameter validation in the wallet-sdk helped a lot to figure this out

mike-parkhill (approved):

PR #247, opened 2022-06-08
fixing issue with token price

maycon-mello:


maycon-mello:
Created a follow up jira for this. https://dock-team.atlassian.net/browse/DCKA-775
We should move the price-service to the wallet-sdk ideally

SamHellawell (approved):

mike-parkhill (approved):

PR #251, opened 2022-06-09
Log exceptions

osinakayah:


src/features/credentials/credentials.js
```diff
@@ -144,22 +145,28 @@ export function getParamsFromUrl(url, param) {
...
-  });
-  if (keyDocs.length > 0) {
-    const subject = {
-      state: getParamsFromUrl(url, 'id'),
-    };
-    const verifiableCredential = await credentialServiceRPC.generateCredential({
-      subject,
+  try {
+    const keyDocs = await wallet.query({
+      type: 'Ed25519VerificationKey2018',
```
SamHellawell, line 150: we need to search for a did document and then find the correlated key from that, a user could have an ed25519 key stored for an account - wont always be a did

SamHellawell: (this can be addressed in separate PR just didnt see this code until now)

osinakayah: @SamHellawell 
Make sense. Right now, when keys for accounts are stored, the type = "KeyPair", so at the moment any key with type = "ed25519" belongs to DID.

SamHellawell: its true for our implementation, but we should consider (for future proofing and going forward) that people may import wallets from other places too where this isnt the case

mike-parkhill (approved):

PR #253, opened 2022-06-13
Fix/testnet transactions

maycon-mello:


mike-parkhill (approved):
We should probably think about how we test for this in the future.

PR #254, opened 2022-06-13
improve error logging

maycon-mello:


src/features/app/app-slice.js
```diff
@@ -144,6 +147,18 @@ export const appOperations = {
   initialize: () => async (dispatch, getState) => {
     SplashScreen.hide();
 
+    const wallet = Wallet.getInstance();
+
+    wallet.eventManager.on(WalletEvents.ready, () => {
```
maycon-mello, line 152: It could potentially fix the issue, in case the app state was not updated properly

mike-parkhill (approved):

PR #259, opened 2022-06-14
Fix/transaction history

maycon-mello:
depends on https://github.com/docknetwork/react-native-sdk/pull/34


src/features/accounts/AccountDetailsScreen.js
```diff
@@ -494,10 +498,55 @@ export function AccountDetailsScreen({
...
+      loadTransactions();
+    },
+    [address, txModule, loadTransactions, setFetching],
+  );
+
+  useEffect(() => {
+    fetch();
+
+    const listener = addr => {
+      console.log('new tx added for', addr);
```
SamHellawell, line 530: best not to keep this in incase it could leak

maycon-mello: fixed

mike-parkhill (approved):

PR #278, opened 2022-06-23
data migration issues

maycon-mello:


src/features/app/app-slice.js
```diff
@@ -145,10 +145,12 @@ export const appOperations = {
     return;
   },
   initialize: () => async (dispatch, getState) => {
-    SplashScreen.hide();
-
     const wallet = Wallet.getInstance();
 
+    wallet.eventManager.on(WalletEvents.migrated, () => {
+      SplashScreen.hide();
```
mike-parkhill, line 151: does the splashscreen get closed elsewhere when no migration is run?

maycon-mello: migration event will happen every time we open the wallet, to check if the current version is the latest. It means the app is ready to be used

maycon-mello: if a migration is required it will just take more time to remove the splash screen, which is expected

maycon-mello:
We need to merge https://github.com/docknetwork/react-native-sdk/pull/40 before merging this one

PR #279, opened 2022-06-23
Test JSON password

osinakayah:


src/features/account-creation/create-account-slice.test.js
```diff
@@ -47,5 +51,29 @@ describe('Create accounts slice', () => {
...
+            json: true,
+          },
+          mnemonicPhrase: '',
+          accountToBackup: null,
+        },
+      });
+      return store
+        .dispatch(createAccountOperations.unlockJson('test'))
+        .then(() => {
+          expect(keyringService.addFromJson).toHaveBeenCalled();
```
mike-parkhill, line 75: will this blow up if the password is wrong? It's not clear to me how knowing addFromJson() was called proves that the password is correct.

osinakayah: Yes, addFromJson tries to decode the json with the password, if that worked, it means the password is correct, if it does not, it means the password was wrong.

There is no explicit method to check if a password is correct or not, so this is just a workaround that achieves the same result.

mike-parkhill: Are we then decoding the JSON twice? Once to test the password and then once when we actually want it? Would it be possible to pass the output from the password check to wherever the JSON is actually needed?

osinakayah: Yes, we will be decoding it twice.
It is possible to pass the output around to the second place the same action will be done again, but I feel we might be tweaking the whole import account flow too much if we do that. 

And I decoding is not an expensive action, we are probably not losing anything  much from doing it twice.

mike-parkhill: Ok, fair enough. It's an optimization we can consider later on if it starts becoming a problem.

PR #296, opened 2022-07-07
fix for "send max token is not working"

maycon-mello:
https://dock-team.atlassian.net/browse/DCKA-887

### DIFF
--- a/src/features/transactions/transactions-slice.js
+++ b/src/features/transactions/transactions-slice.js
@@ -159,7 +159,14 @@ export const transactionsOperations = {
     },
 
   sendTransaction:
-    ({recipientAddress, accountAddress, amount, fee, prevTransaction}) =>
+    ({
+      recipientAddress,
+      accountAddress,
+      amount,
+      fee,
+      prevTransaction,
+      sendMax,
+    }) =>
     async (dispatch, getState) => {
       showToast({
         message: translate('send_token.transaction_sent'),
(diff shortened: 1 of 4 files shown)


src/features/transactions/transactions-slice.js
```diff
@@ -183,6 +190,7 @@ export const transactionsOperations = {
         .sendTokens({
           toAddress: recipientAddress,
           fromAddress: accountAddress,
+          transferAll: sendMax,
```
mike-parkhill, line 193: what's this used for?

mike-parkhill: do we need unit test updates to reflect this change?

maycon-mello: there is no test for this yet, worth adding integration tests

maycon-mello:
> side comment but noticed in this PR. why are we committing the bundle JS? shouldnt it be ignored in favour of building it from source?

I agree we should be able to add it as part of ci, let me create a jira for it

maycon-mello:
There is a command already, `yarn build-sdk` it will generate the bundle JS, should be a quick fix actually

maycon-mello:
> There is a command already, `yarn build-sdk` it will generate the bundle JS, should be a quick fix actually

since we have merged the fix already I did a rebase on this one, so that we are not pushing the bundle anymore

SamHellawell (approved):

PR #304, opened 2022-07-22
Import DID UI

osinakayah:


src/features/didManagement/didHooks.js
```diff
@@ -20,6 +22,18 @@ export function useDIDManagementHandlers() {
     _hasError: false,
   });
 
+  const onImportDID = useCallback(
+    async ({encryptedJSONWallet, password}) => {
+      await importDID({encryptedJSONWallet, password});
```
mike-parkhill, line 27: how are we handling errors from `importDID`?

src/features/didManagement/didHooks.test.js
```diff
@@ -78,4 +78,98 @@ describe('DID hooks', () => {
...
+    const password = 'test';
+    await result.current.onImportDID({encryptedJSONWallet, password});
+
+    const {result: dIDManagementResult} = renderHook(() => useDIDManagement());
+    expect(dIDManagementResult.current.importDID).toBeCalledWith({
+      encryptedJSONWallet,
+      password,
+    });
+  });
+  test('Import DID with invalid params', async () => {
```
mike-parkhill, line 130: can we get a test to see it how handles empty and/or corrupted JSON?

mike-parkhill (commented):
What's the plan for importing by QR code scan?

src/features/didManagement/didHooks.js
osinakayah: https://github.com/docknetwork/dock-app/blob/4a129c4c9fda67bc2658effde4e6a964508f44cb/src/features/didManagement/didHooks.js#L92


onImportDID did is wrapped with "withErrorToast" function. This function handles the exception if it happens.
withErrorToast first displays the error as a toast, then logs the error on sentry and google analytics.

src/features/didManagement/didHooks.test.js
osinakayah: It can be tested here via the didHooks because the function was mocked, but has been tested here

https://github.com/docknetwork/react-native-sdk/blob/2bdc87d57c12ed7afb3c3f68cca243c2ec7b8978/packages/core/lib/services/wallet/index.test.js#L225

osinakayah:
> What's the plan for importing by QR code scan?

We are discussing that. We are considering compressing the the DID so it can encoded in a QRCode, we still discussing it though.

maycon-mello (approved):

PR #322, opened 2022-08-16
fix issue with export did

maycon-mello:


osinakayah:
@maycon-mello 
Some tests are failing

osinakayah (approved):

PR #328, opened 2022-08-24
Fix import invalid Credentails

osinakayah:


src/features/credentials/credentials.js
```diff
@@ -116,6 +116,11 @@ export function useCredentials({onPickFile = pickJSONFile} = {}) {
       return;
     }
 
+    assert(
+      jsonData?.issuer?.hasOwnProperty('id') === true,
```
mike-parkhill, line 120: is `id` the only mandatory field, or should we assert others too?

src/features/credentials/credentials.test.js
```diff
@@ -179,10 +179,35 @@ describe('Credentials helpers', () => {
       expect(Credentials.getInstance().remove).toBeCalledWith(1);
     });
 
-    it('expect to add credential', async () => {
-      await act(async () => {
-        await hook.current.onAdd();
+    it('expect to add invalid credential', async () => {
```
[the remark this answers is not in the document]
osinakayah: It is not adding a credential, the test is is making sure that an exception is thrown. The test description has been changed now now

src/features/credentials/credentials.js
osinakayah: issuer.id is not, other mandatory fields are  id, @context,type too.  Other wallet documents (DIDsDocument, WalletBackup) has this properties too, except "issuer.id".

src/features/credentials/credentials.test.js
```diff
@@ -179,7 +179,7 @@ describe('Credentials helpers', () => {
       expect(Credentials.getInstance().remove).toBeCalledWith(1);
     });
 
-    it('expect to add invalid credential', async () => {
+    it('expect adding invalid credential to throw exception', async () => {
```
mike-parkhill, line 182: should we test the various invalid scenarios?

osinakayah: Yes, we can do that.

src/features/credentials/credentials.js
```diff
@@ -89,24 +89,24 @@ export async function processCredential(credential) {
 const validateCredential = credential => {
   assert(
     typeof credential !== 'undefined',
-    translate('credentials.invalid_credential') + 0,
+    translate('credentials.invalid_credential'),
```
mike-parkhill, line 92: I kind of liked having some way of knowing what was wrong with the credential based on a code or something. Could we still have that?

osinakayah: Yes, We can have a different  error message based on the mandatory property that is missing.

mike-parkhill (approved):

PR #330, opened 2022-08-26
Dcka 975 presentation

osinakayah:


src/core/hooks/deepLinkHooks.test.js
```diff
@@ -0,0 +1,19 @@
+import {renderHook, act} from '@testing-library/react-hooks';
+import {useDeepLink} from './deepLinkHooks';
+import {Routes} from '../routes';
+import {navigate} from '../navigation';
+
+describe('Deep hooks', () => {
+  it('should navigate to presentation screen', () => {
```
mike-parkhill, line 7: can you add a test case for the auth flow?

src/features/credentials/components/SingleCredentialToShare.js
```diff
@@ -0,0 +1,89 @@
...
+import {NBox, Theme, Typography} from '../../../design-system';
+import {View} from 'react-native';
+import {PolkadotIcon} from '../../../components/PolkadotIcon';
+import {getDIDAddress} from '../credentials';
+import {formatDate} from '@docknetwork/wallet-sdk-core/lib/core/format-utils';
+import {withErrorBoundary} from '../../../core/error-handler';
+import {translate} from '../../../locales';
+import {renderObjectAttributes} from '../CredentialsScreen';
+
+export const SingleCredentialToShare = withErrorBoundary(
```
mike-parkhill, line 12: is this a different view than the one used in the general credential list? Could they share one?

src/features/qr-code-scanner/qr-code.js
```diff
@@ -168,6 +169,11 @@ export async function qrCodeHandler(data, handlers = qrCodeHandlers) {
 }
 
 export function isDidAuthUrl(url) {
```
mike-parkhill, line 171: is having this helper function overly useful now that you have the generic one below?

src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -91,28 +95,24 @@ describe('qr-code', () => {
         .spyOn(utilCryptoService, 'isAddressValid')
         .mockReturnValueOnce(false);
 
-      navigationRef.current = {
-        navigate: jest.fn(),
-      };
+      // navigationRef.current = {
```
mike-parkhill, line 98: commented code?

src/core/hooks/deepLinkHooks.test.js
osinakayah: We can capture that with an integration test better.

src/features/credentials/components/SingleCredentialToShare.js
osinakayah: The UI is slightly different, the current UI has a check button for selecting and sharing selected credentials.

src/features/qr-code-scanner/qr-code.js
osinakayah: If I change the function name it would require a little bit of more refactoring, so I opted to change the body of the function instead, that way the logic is still not been duplicated.

src/features/credentials/components/SingleCredentialToShare.js
mike-parkhill: I could see us supporting multi-select on the main list at some point (e.g for multiple delete, etc.) in which case they'd be the same. If the only difference is support of multi-select I'd use the same view and have multi-select support be configurable

advantages being that they'll always look the same instead of having to worry about syncing them any time we make changes

src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -252,7 +249,24 @@ describe('qr-code', () => {
 
       expect(credentialServiceRPC.signCredential).toBeCalled();
     });
-
+    it('is Presentation QRCode scanned', () => {
+      const isValid1 = onPresentationScanned(
+        'dockwallet://proof-request?ul=https://auth.dock.io/verify?id=dockstagingtestRgMV0IwPQELYDbVkGXUfMQnOb912660w&scope=public email',
```
mike-parkhill, line 254: can we make it more obvious that the scanned link is invalid? My first response in reading this was to fix the query parameter `ul` to be '`url`, but then realized it was probably meant to be wrong.

src/features/credentials/components/SingleCredentialToShare.js
osinakayah: True. 
Though, another difference in the UI is a Kebab option menu that pops up a modal for deleting a single credentials.

mike-parkhill: can we pass the actions supported by the particular instance of the view to the view? Or, have the containing list handle the actions based on the needs of the current context?

In UI in general (not just react), It feels wrong to me to have to recreate the same view everywhere you want to display the same data just because the actions change based on the context. The data to display is the same, so the view should be the same. The actions should be part of the context - so either injected into the view or the container around the view listens for events on the view and handles it appropriately.

mike-parkhill (approved):

PR #332, opened 2022-08-29
Dcka 894 import wrong credential

osinakayah:


src/features/credentials/CredentialsScreen.js
```diff
@@ -154,11 +154,28 @@ const CredentialListItem = withErrorBoundary(
                   }}
                 />
               </View>
-            ) : (
+            ) : getDIDAddress(credential.issuer) ? (
```
mike-parkhill, line 157: how expensive is this call? We're doing it twice here.

src/features/credentials/CredentialsScreen.test.js
```diff
@@ -19,4 +19,81 @@ describe('CredentialsScreen', () => {
     );
     expect(wrapper.dive()).toMatchSnapshot();
   });
+  it('should render correctly credential list', () => {
```
mike-parkhill, line 22: there are two paths through the modified code, should there be at least two tests?

src/features/credentials/CredentialsScreen.js
osinakayah: It is not expensive. Basically it just gets a substring from a string if certain criteria are met, if not, returns null.

PR #333, opened 2022-08-30
DCKA-905: fix issue with web3-id signin

maycon-mello:


mike-parkhill (commented):
any tests we can write to verify this is working as intended?

mike-parkhill (approved):

PR #335, opened 2022-08-31
Presentation handler

osinakayah:


src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -324,5 +325,9 @@ describe('qr-code', () => {
         },
       );
     });
+    it('Include presentation QR handlers', () => {
+      expect(qrCodeHandlers.length).toBe(4);
+      expect(qrCodeHandlers[3]).toBe(onPresentationScanned);
```
mike-parkhill, line 330: would it be safer to just check for existence in the array rather a specific position - would make the test a little more flexible and I don't think the specific order matters, does it?

mike-parkhill (approved):

PR #338, opened 2022-09-02
Improvements

osinakayah:


src/features/credentials/components/SelectDIDComponent.js
```diff
@@ -9,36 +9,38 @@ export function SelectDIDComponent({dids, handleChange}) {
...
-          />
-        </Stack>
+        <Typography variant="h1" mb={12}>
+          {translate('credentials.select_presentation_holder')}
+        </Typography>
+
+        <FormControl.Label>
+          {translate('didManagement.did_name')}
+        </FormControl.Label>
+        <CustomSelectInput
```
mike-parkhill, line 19: would it make sense to display the single DID here as a label rather than a drop down if there's only one? 

or is the thinking that there's no value even showing the DID if we're auto-selecting it?

src/features/credentials/hooks/credentialPresentation.js
```diff
@@ -76,18 +76,38 @@ export function useCredentialPresentation(deepLinkUrl) {
...
       step,
-      onNext: onNext,
+      onNext: withErrorToast(onNext),
+      onPresentCredentials: withErrorToast(onPresentCredentials),
       onSelectDID,
       isFormValid,
     };
   }, [isFormValid, onNext, onPresentCredentials, selectedCredentials, step]);
 }
+export function useSingleDID(dids, onSelectDID) {
```
mike-parkhill, line 107: I think I'd expect a DID function to be in the DID component, not here

src/locales/languages/en.json
```diff
@@ -300,6 +300,7 @@
     "empty_items": "You’ll see credentials here once you accept them",
     "default_title": "Untitled Credential",
     "select_what_to_share": "Select what data to share",
+    "select_presentation_holder": "Select Presentation holder",
```
mike-parkhill, line 303: not sure this text will mean much to the average user who probably doesn't know what a Holder or Presentation is.

What about something like, "Select identity to present" or something?

src/features/credentials/components/SelectDIDComponent.js
osinakayah: Right now, if there is a single DID, this screen does not get displayed, we just select the DID and use it as the identity.

mike-parkhill (approved):

PR #339, opened 2022-09-02
Credential validity

osinakayah:


src/features/credentials/credentials.js
```diff
@@ -106,6 +106,10 @@ const validateCredential = credential => {
     credential.type?.includes('VerifiableCredential'),
     translate('credentials.credential_no_type'),
   );
+  assert(
+    !credential.type?.includes('EncryptedWallet'),
```
mike-parkhill, line 110: this feels like we'll continually be adding new types that don't here as we discover them. Is there some way we can specifically look for what we know we support (i.e. Verifiable Credential) rather than an exclusion of the infinite types we don't?

osinakayah: I discussed this with @SamHellawell.
https://dock-team.slack.com/archives/C03PEPUB62D/p1662137893368659?thread_ts=1662109221.562899&cid=C03PEPUB62D

It might be difficult to accurate target the type of credential that can be imported and also future proof it.

mike-parkhill: ok 😢

mike-parkhill (approved):

maycon-mello (approved):

PR #342, opened 2022-09-09
QR Handler

osinakayah:


src/core/storage-utils.test.js
```diff
@@ -36,4 +36,13 @@ describe('storage utils', () => {
 
     expect(err.code).toBe('OTHER_CODE');
   });
+
+  it('parse string to JSON', () => {
+    const res = stringToJSON('dockwallet://proof-request?url=https%3A%2F%2Fap');
```
maycon-mello, line 41: How would we detect the json string is invalid? Perhaps we should return a `null` instead?

mike-parkhill: or just let the error bubble out and expect the caller to handle it

or go with a tryParse() model or something

osinakayah: stringToJSON is meant to accept a string and return an JSON object.
if it can't parse the string to a JSON object, it returns an empty JSON object, not null, that way, the function stays true to its return type. It is just a personal preference though.

> How would we detect the json string is invalid
We can create another helper function, that returns true/false if the string is a valid JSON string.

mike-parkhill: I guess the question is whether returning `{}` could ever be misleading. '{}' is legitimate JSON, but `{I'mnot JSON >>>` is clearly not but would provide the same output. Checking for null or allowing the exception to bubble out makes it more explicit that the input was invalid.

mike-parkhill (approved):

PR #343, opened 2022-09-12
Remove realm files

osinakayah:


mike-parkhill (commented):
this is just cleaning up old code that's moved down to the SDK?

maycon-mello (approved):

PR #344, opened 2022-09-13
verifier flow

maycon-mello:


mike-parkhill (commented):
does this actually run the verification?

src/features/credentials/hooks/credentialPresentation.js
```diff
@@ -8,11 +8,74 @@ import {navigate} from '../../../core/navigation';
...
+async function handleQRCodePresentation({
+  parsedSelectedCredentials,
+  keyDoc,
+  presentCredentials,
+  setPresentationData,
+}) {
+  const presentation = await presentCredentials({
+    credentials: parsedSelectedCredentials,
+    keyDoc,
+    challenge: `${Date.now()}`,
```
mike-parkhill, line 63: should probably add something random to this to ensure uniqueness

mike-parkhill (commented):
Still have the one comment about the request challenge and it looks like this PR is set against `main` instead of `dev`

mike-parkhill (approved):

PR #345, opened 2022-09-13
Separated navigation stacks

osinakayah:


src/features/wallet/wallet-slice.js
```diff
@@ -250,11 +250,12 @@ export const walletOperations = {
         }
       }
       dispatch(authenticationActions.setAuth({isLoggedIn: true}));
+
       if (callback) {
         callback();
       } else {
         dispatch(accountOperations.loadAccounts());
-        navigate(Routes.ACCOUNTS);
+        // navigate(Routes.ACCOUNTS);
```
mike-parkhill, line 258: is this intended to be commented out?

src/features/wallet/wallet-slice.js
```diff
@@ -308,6 +309,7 @@ export const walletOperations = {
 
       dispatch(walletActions.setCreationFlags({}));
 
+      // TODO change auth state
```
mike-parkhill, line 312: while you're in here does it make sense to start thinking about how we'll handle it when Accounts is turned off with a feature flag?

i.e. instead of going to Routes.ACCOUNTS, should we put the active options into an array and then just grab the first one or something?

src/features/wallet/wallet-slice.js
osinakayah: Yes, we won't have to manually navigate to the Accounts screen anymore. The comment will be removed.

src/features/wallet/wallet-slice.js
osinakayah: I will still make changes here, because I will be changing the accounts, credentials, dids and settings routes from a stack navigator to a tab navigator.
After this changes are implemented, I will then load the tabs from a config array or something like that.

mike-parkhill (approved):

PR #346, opened 2022-09-15
Using Tab navigator

osinakayah:


src/core/navigation/authenticated-stack.js
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 import {createStackNavigator} from '@react-navigation/stack';
+import {createBottomTabNavigator} from '@react-navigation/bottom-tabs';
```
mike-parkhill, line 3: Any opportunities to split this file up? It's pretty big

PR #353, opened 2022-09-26
Credential

osinakayah:


src/features/credentials/credentials.test.js
```diff
@@ -309,10 +230,20 @@ describe('Credentials helpers', () => {
       const {result} = await renderHook(() =>
         useCredentials({onPickFile: onPickValidFile}),
       );
+      const {result: useCredentialUtilsResult} = await renderHook(() =>
+        useCredentialUtils({onPickFile: {}}),
```
maycon-mello, line 234: it seems like `onPickFile` should be a function instead of object

src/features/credentials/credentials.test.js
```diff
@@ -163,39 +104,19 @@ describe('Credentials helpers', () => {
...
-
-    it('expect to sort credentials', () => {
-      expect(creds.sort).toBeCalled();
-    });
-
     it('expect to delete credential', async () => {
-      await act(() => {
-        hook.current.handleRemove({id: 1});
-      });
+      const {result} = await renderHook(() => useCredentials({onPickFile: {}}));
```
maycon-mello, line 111: it seems like `onPickFile` should be a function instead of object

src/features/credentials/credentials.test.js
```diff
@@ -163,39 +104,19 @@ describe('Credentials helpers', () => {
...
+      const {result: useCredentialUtilsResult} = await renderHook(() =>
+        useCredentialUtils({onPickFile: {}}),
```
maycon-mello, line 113: same regarding `onPickFile`

src/features/credentials/credentials.js
```diff
@@ -121,25 +79,22 @@ const validateCredential = credential => {
...
+  const {credentials, saveCredential, deleteCredential} = useCredentialUtils();
   const [items, setItems] = useState([]);
 
-  const syncCredentials = async () => {
-    const credentials = await Credentials.getInstance().query();
-
+  const syncCredentials = useCallback(async () => {
     const processedCredentials = await Promise.all(
-      credentials.sort(sortByIssuanceDate).map(processCredential),
+      credentials.map(processCredential),
```
maycon-mello, line 87: What if we rename `processCredential` function helper to something more specific. Maybe `formatCredential`? Same for `processedCredentials` const

maycon-mello (changes requested):
small requests

maycon-mello (approved):

PR #354, opened 2022-09-27
chore: add test ids on account list

maycon-mello:


mike-parkhill (commented):
no snaps needed updating?

PR #358, opened 2022-09-30
Config more colors

osinakayah:


src/components/__snapshots__/PasswordInputScreen.test.js.snap
```diff
@@ -33,7 +33,7 @@ exports[`PasswordInputScreen should render correctly 1`] = `
         <Memo(ForwardRef(Box))
           mt={7}
         >
-          <Memo(ForwardRef(Input))
```
mike-parkhill, line 36: what's causing this change?

osinakayah: I added a new prop to the input fields so that the background color of the input fields can be configured from the themes styles.

mike-parkhill (approved):

PR #368, opened 2022-10-07
removing advanced options for DIDs

maycon-mello:


src/features/didManagement/CreateNewDIDScreen.js
```diff
@@ -21,7 +21,6 @@ import {
   VStack,
 } from 'native-base';
 import {translate} from '../../locales';
-import {DIDAdvancedOptions} from './components/DIDAdvancedOptions';
```
mike-parkhill, line 24: should this component file be deleted too?

mike-parkhill (commented):
Will the default be set properly with this block gone?

I imagine there might need to be matching updates in the the automated UI tests

maycon-mello:
> Will the default be set properly with this block gone?
> 
> I imagine there might need to be matching updates in the the automated UI tests

Yes, this block was optional

src/features/didManagement/CreateNewDIDScreen.js
maycon-mello: The component was deleted in the latest commit

PR #377, opened 2022-10-17
use packaging.config to generate packages

maycon-mello:


ios/DockApp.xcodeproj/project.pbxproj
```diff
@@ -801,9 +801,9 @@
 					"-ObjC",
 					"-lc++",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = labs.dock.DockWalletUI;
+				PRODUCT_BUNDLE_IDENTIFIER = labs.dock.DockApp;
```
mike-parkhill, line 804: is this something we'll need to update for the white label clients?

mike-parkhill: nvm, I see you're doing that with a sed command above

mike-parkhill (approved):

PR #383, opened 2022-10-25
Warning before importing expired credential

osinakayah:


src/features/credentials/credentials.js
```diff
@@ -99,12 +101,30 @@ export function useCredentials({onPickFile = pickJSONFile} = {}) {
...
     if (!jsonData) {
       return;
     }
     validateCredential(jsonData);
-    await saveCredential(jsonData);
+    const hasExpired = jsonData.expirationDate
+      ? isInThePast(new Date(jsonData.expirationDate))
+      : false;
+
+    if (hasExpired) {
```
maycon-mello, line 112: We could check for non expired cred first
```js
if (!hasExpired) {
   return saveCredential(jsonData);
}

// show confirmation modal
```

src/features/credentials/credentials.js
```diff
@@ -99,12 +101,30 @@ export function useCredentials({onPickFile = pickJSONFile} = {}) {
 
   const onAdd = async () => {
```
maycon-mello, line 102: is there any unit test update for `onAdd`?

PR #384, opened 2022-10-25
Include dock balance

osinakayah:


src/features/didManagement/components/SingleDIDCreationPaymentAccount.test.js
```diff
@@ -0,0 +1,23 @@
...
+const mockStore = configureMockStore();
+
+describe('SingleDIDCreationPaymentAccount', () => {
+  it('should render correctly ', () => {
+    const initialState = {
+      // placeholder for redux store
+    };
+
+    const wrapper = shallow(
+      renderAppProviders(<SingleDIDCreationPaymentAccount item={{}} />),
```
mike-parkhill, line 16: should we test it with an item that looks like an account?

mike-parkhill (approved):

PR #395, opened 2022-11-10
Config firebase

osinakayah:


mike-parkhill (commented):
Where did this change come from?

osinakayah:
> Where did this change come from?

This are the changes we need to make for firebase to work. 
I think they were removed when we upgraded react native

mike-parkhill (approved):

PR #397, opened 2022-11-10
fix: issue with import modal

maycon-mello:
https://dock-team.atlassian.net/browse/DCKA-1138


mike-parkhill (approved):
can we add a snapshot test or something?

maycon-mello:
> can we add a snapshot test or something?

This was related to an internal state issue. Maybe we should catch this on e2e testing?

PR #406, opened 2022-11-17
fix: feature flag verifier feature

maycon-mello:


osinakayah (changes requested):
Can we update the snapshot to reflect this change?

osinakayah (approved):

PR #420, opened 2022-11-22
fix/qr code crashing

maycon-mello:
- fix: qr code crashing
- fix: qr code crashing


osinakayah (changes requested):
Can we update the snapshot? 
For when the "data" is undefined and for when it is defined.

osinakayah (approved):

osinakayah (approved):

PR #429, opened 2022-11-29
fix: issues with did fragment

maycon-mello:


src/features/qr-code-scanner/qr-code.test.js
```diff
@@ -450,6 +451,37 @@ describe('qr-code', () => {
       expect(qrCodeHandlers.includes(onPresentationScanned)).toBeTruthy();
     });
   });
+
+  describe('ensureDIDDockFragment', () => {
```
SamHellawell, line 455: these tests are good but is https://github.com/docknetwork/dock-app/pull/429/files#diff-fc7667b2f5447459bbdd39e3cce647af4df72c17fad13e00e277ef4d7cff7271R192 tested for each scenario/did type?

maycon-mello: Let me move this implementation to the wallet-sdk repository, so that we can add proper tests to it

SamHellawell (approved):
approving now so we can merge

PR #447, opened 2022-12-20
Feat/pex proof request flow

maycon-mello:


SamHellawell (approved):
looks good for this update, but just checking in general if legacy QR code requests are still supported? we will phase them out soon but not immedtiately due to people using them

src/features/qr-code-scanner/qr-code.js
```diff
@@ -312,7 +313,33 @@ export async function relayServiceHandler(json) {
   }
 }
 
+export async function pexMessageHandler(value) {
```
mike-parkhill, line 316: this file's starting to get big - should we consider starting to break the handlers out into their own files?

maycon-mello: yes, I think we can also improve performance. Each handler perform a JSON.parse, we can do it once and pass along for all the handlers, same for fetching data from URL. Let me create a jira to perform this properly

SamHellawell: was thinking the same but would argue its better to refactor in later PR (unless its super quick to do) - just want to land this so i can finish up the certs side and have the team test it too

maycon-mello: https://dock-team.atlassian.net/browse/DCKA-1251

maycon-mello:
> looks good for this update, but just checking in general if legacy QR code requests are still supported? we will phase them out soon but not immedtiately due to people using them

Yes, legacy flow is still working

src/features/qr-code-scanner/qr-code.js
mike-parkhill: Yep, I'm good with doing it later, just wanted to call it out

PR #459, opened 2023-01-19
fix: issues with credential filter

maycon-mello:
I've found specific implementation of the credential filter. I've removed it to fix some performance issues, and duplicated requests to pex filters.


src/features/credentials/hooks/credentialPresentation.js
```diff
@@ -11,7 +11,11 @@ import {translate} from '../../../locales';
 import {navigate} from '../../../core/navigation';
 import axios from '../../../core/network-service';
 import {Routes} from '../../../core/routes';
-import {formatCredential, isBBSPlusCredential} from '../credentials';
+import {
+  formatCredential,
```
mike-parkhill, line 15: I'm not seeing where `formatCredential` and `isBBSPlusCredential` are being used

src/features/credentials/hooks/credentialPresentation.js
```diff
@@ -209,8 +207,7 @@ export function useCredentialPresentation({
 }
 
 export function useBBSCredentialPresentation({presentationDefinition}) {
```
SamHellawell, line 209: curious - why is this method needed/what does it do? not sure of side effects of this change

for me it seems odd that BBS is taking a separate path, the wallet should consider them the same as any other VC but when choosing that one to present you can also choose what attributes to present/auto select them based off of inputs. perhaps im just not understanding what this method is used for though

SamHellawell (approved):
code lgtm but im not aware of side effect changes of the PR, made a comment on one method to try understand more

mike-parkhill (approved):

PR #461, opened 2023-01-20
Enforce to enter password

osinakayah:
Remove the password parameter from url to enforce password entry.


SamHellawell:
this PR works but https://github.com/docknetwork/react-native-sdk/blob/master/packages/credentials/lib/index.js#L78 should be updated, instead of fetching and expecting 400 then it should check if url is [creds.dock.io](http://creds.dock.io/) or [creds-testnet.dock.io](http://creds-testnet.dock.io/) as this auth mechanism is for dock's ecosystem only + saves a non-needed request

src/features/qr-code-scanner/qr-code.js
```diff
@@ -55,22 +55,18 @@ export async function credentialHandler(data) {
           type: 'message',
           message: translate('global.fetching_data'),
         });
-        const isPasswordNeeded = await credentials.isPasswordRequired(data);
```
SamHellawell, line 58: this check should be re-added `credentials.isPasswordRequired` - and ideally that should check that its a valid dock certs URL. eg:
- https://creds.dock.io
- https://creds-testnet.dock.io
- https://creds-staging.dock.io

if url doesnt start with these, then skip all password logic

src/features/qr-code-scanner/qr-code.js
```diff
@@ -58,8 +58,9 @@ export async function credentialHandler(data) {
...
-        if (isPasswordNeeded) {
+        const isDockCertURL = credentials.isDockCertsURL(url);
+        const url = removeParam('p', data);
```
SamHellawell, line 62: param should only be removed if its a dock certs url

SamHellawell:
looks good now just unit tests need fixing

PR #464, opened 2023-01-23
fix: prevent duplicated calls on verification submission

maycon-mello:


src/features/credentials/SelectBBSHolderDIDToShareScreen.js
```diff
@@ -39,7 +40,7 @@ export function SelectBBSHolderDIDToShareScreen({
               onPress={onDecline}>
               {translate('navigation.decline')}
             </Button>
-            <Button
+            <LoadingButton
```
mike-parkhill, line 43: so this prevents a duplicate request caused by double clicks or whatever, what about the scenario where the same QR code is used in separate flows?

maycon-mello: There will be an error response from certs in this case

maycon-mello: image

mike-parkhill (dismissed):

mike-parkhill (approved):

PR #466, opened 2023-01-25
DCKA 1342 trim text length

maycon-mello:
DCKA 1342 trim text length


src/features/credentials/components/CredentialAttributesComponent.js
```diff
@@ -33,10 +33,12 @@ export function SingleCredentialAttribute({attribute, actions = <NBox />}) {
         {attribute.name}
       </Typography>
       <HStack>
-        <Typography flexGrow={1} variant={'description'} mb={1}>
-          {attribute.value}
-        </Typography>
-        {actions}
+        <Box flex={1}>
```
mike-parkhill, line 36: how does this do the trim? Is it built in to the component?

maycon-mello: I've fixed the layout to wrap the value without breaking the checkbox component. In that case trimming a DID would not be required. But thinking o blobs and bigger data, I've now added a trim function to ensure those values should be handled properly. Please check
image

mike-parkhill (dismissed):

mike-parkhill (approved):

PR #472, opened 2023-01-27
fix/qr code navigation issues

maycon-mello:
fix/qr code navigation issues


mike-parkhill (approved):
no snapshots needed updating?

maycon-mello:
> no snapshots needed updating?

there are no tests for the OptionList component, I can push a separate PR for that. should be quick

PR #483, opened 2023-02-13
removing feature "share presentation qr code"

maycon-mello:
dev


mike-parkhill (approved):
Is there dead code we need to come back to for clean-up still?

maycon-mello:
> Is there dead code we need to come back to for clean-up still?

Yes, there is a PresentationFlow.qrCode in the presentation handler to be removed still. I've added a note on https://dock-team.atlassian.net/browse/DCKA-1336

PR #484, opened 2023-02-14
feat/improve logging

maycon-mello:
dev


src/core/events.ts
```diff
@@ -14,11 +15,14 @@ export const enum EventTypes {
 export class AppEventEmitter extends EventEmitter {
   emit(eventName: EventTypes, ...args: any[]) {
     assert(!!eventName, 'Event name is required');
+    Logger.debug(`Event emitted: ${eventName}`);
+    Logger.debug(`Event args: ${JSON.stringify(args)}`);
```
mike-parkhill, line 19: any chance of PII in `args`?

src/core/logger.js
```diff
@@ -1,8 +1,18 @@
-import {LOGGER_LEVEL} from '@env';
 const winston = require('winston');
+const debugLogs = [];
 
 export const Logger = winston.createLogger({
   format: winston.format.simple(),
-  level: LOGGER_LEVEL || 'info',
+  level: 'debug',
```
mike-parkhill, line 6: any value in allow this to be configurable in Settings?

src/features/dev-settings/hooks/requestLogHooks.js
```diff
@@ -4,17 +4,19 @@ import RNFS from 'react-native-fs';
 import {exportFile} from '../../accounts/account-slice';
 import {translate} from '../../../locales';
 import {showToast, withErrorToast} from '../../../core/toast';
+import {getDebugLogs} from 'src/core/logger';
 
 export function useRequestLogger() {
   const exportLogRequest = useCallback(async () => {
     const logs = RequestLogger.exportLog();
+    const debugLogs = getDebugLogs() || [];
```
mike-parkhill, line 12: would a simple viewer be useful in the app? Might make it easier for us when debugging something.

not saying it's needed for this PR, but might be useful to add at some point

maycon-mello: good point, can allow the user to copy the data to clipboard. This would be useful for emulators too

src/core/events.ts
maycon-mello: yeah, its better not to print the args

src/features/qr-code-scanner/qr-code.js
```diff
@@ -205,19 +204,25 @@ export async function onPresentationDefinitionScanned(value) {
         presentationDefinition:
           typeof data === 'string' ? JSON.parse(data) : data,
         onSubmitPresentation: presentationData => {
+          Logger.debug('Submitint presentation data');
+          Logger.debug(JSON.stringify(presentationData));
```
mike-parkhill, line 208: this will almost certainly have PII in it which if the logs stays on the device isn't a problem, but as soon as it gets exported/copied out it could be

maycon-mello: it's meant for developer purposes and is only enabled by feature flag on dev settings, a regular wallet will not log this information. is that fine in that case?

mike-parkhill: yeah, that's probably fine for now.

mike-parkhill (approved):

PR #485, opened 2023-02-15
feat/notifications poc

maycon-mello:
dev


src/core/types.ts
```diff
@@ -0,0 +1,11 @@
+export interface KeyPairDocument {
+  '@context': string[];
+  id: string;
+  controller: string;
+  type: string;
+  publicKeyBase58: string;
```
mike-parkhill, line 6: will there ever be more keybase pairs than the Base58 and Multibase options? Just wondering if a dictionary would be more extensible for adding other options in the future.

maycon-mello: This is the current format we received from dock sdk. In the future we could extend this interface if required, setting `publicKeyBase58` as optional and adding a new other keyBase attribute, or even adding dictionary as you suggested.

maycon-mello: We receive this data from the dock sdk. The dictionary would need to be supported by the sdk when signing messages with a did. Also we should be able to extend this interface in the future, if required.

mike-parkhill (approved):

src/core/types.ts
SamHellawell: @mike-parkhill multibase is intended to cover all encoding cases

src/core/use-notifications.ts
```diff
@@ -0,0 +1,113 @@
...
+      return;
+    }
+
+    debug('Registering device token for wallet DIDs');
+    debug(`Device token: ${deviceToken}`);
+    debug(
+      `Wallet DIDs: ${didDocs.map((didDoc: KeyPairDocument) => didDoc.id)}`,
+    );
+
+    relayService.registerDIDPushNotification({
```
SamHellawell, line 103: can we add some check here to make sure token has changed since the initial generation before submitting request to the relay?

maycon-mello: Sure

maycon-mello: I will push as separate PR, including some other performance improvements

PR #488, opened 2023-02-21
feat/webview sandbox

maycon-mello:
Depends on https://github.com/docknetwork/react-native-sdk/pull/99


mike-parkhill (approved):

mike-parkhill:
Quick sanity check - if someone injected the appropriate malicious code into their filter could they still leak out the VC details that are being passed for filtering? Or is the sandbox locked down to prevent it from reaching out?

PR #493, opened 2023-03-02
feat/improvements for device token registration

maycon-mello:
- fixing issues with duplicated token registration
- call relay service registration only if token/didList has changed


src/core/notifications/register-device-token.ts
```diff
@@ -0,0 +1,89 @@
...
+}: {
+  didDocs: KeyPairDocument[];
+  deviceToken: string;
+  relayService: RelayServiceRpc;
+}) {
+  const didList = didDocs.map((didDoc: KeyPairDocument) => didDoc.id);
+
+  debug('Registering device token for wallet DIDs');
+  debug(`Device token: ${deviceToken}`);
+  debug(`Wallet DIDs: ${didList}`);
```
SamHellawell, line 69: does this debug function have any effect in live apps? if so best to disable leaking and logging of dids

SamHellawell (approved):
minor comment but lgtm

src/core/notifications/register-device-token.ts
maycon-mello: It will have effect only when a feature flag is enabled (via dev settings). By default it will not log anything

mike-parkhill: we should still be careful of what we log - I suggest we don't log private DID keys, passwords, etc. ever

SamHellawell: if a user has to explicitly enable it then i think its ok for now but definitely id remove this in future for token/did logging just incase. while its in dev i can see it being useful

PR #505, opened 2023-03-14
DCKA 1494 hide human unreadable fields

maycon-mello:
https://dock-team.atlassian.net/browse/DCKA-1494


src/features/credentials/hooks/usePrettyVCData.ts
```diff
@@ -80,8 +80,24 @@ const applyCustomDateFormat = ({vcData}: CustomMappingProps): VCData => {
   return vcData;
 };
 
+const MAX_ATTR_LENGTH = 256;
+
+const hideHummanUnreadableFields = ({vcData}: CustomMappingProps): VCData => {
+  return {
+    ...vcData,
+    attributes: vcData.attributes.filter(attr => {
+      if (typeof attr.value === 'string' && attr.value.length > MAX_ATTR_LENGTH) {
```
SamHellawell, line 89: 256 chars seems alot, maybe 128?

another check you can do is checking if its base64 or not, if so then we can hide it

src/features/credentials/hooks/usePrettyVCData.ts
```diff
@@ -80,8 +80,35 @@ const applyCustomDateFormat = ({vcData}: CustomMappingProps): VCData => {
...
+const MAX_ATTR_LENGTH = 128;
+
+const base64Regex = /^(?:[A-Za-z0-9+/]{4})*?(?:[A-Za-z0-9+/]{2}(?:==)?|[A-Za-z0-9+/]{3}=?)$/;
+// TODO: use wallet-sdk base64 util
+// Will need json-rpc method for that
+function isBase64(value: string) {
+  return base64Regex.test(value);
+}
+
+const hideHummanUnreadableFields = ({vcData}: CustomMappingProps): VCData => {
```
mike-parkhill, line 92: typo: `Humman`

src/features/credentials/hooks/usePrettyVCData.ts
```diff
@@ -80,8 +80,35 @@ const applyCustomDateFormat = ({vcData}: CustomMappingProps): VCData => {
   return vcData;
 };
 
+const MAX_ATTR_LENGTH = 128;
+
+const base64Regex = /^(?:[A-Za-z0-9+/]{4})*?(?:[A-Za-z0-9+/]{2}(?:==)?|[A-Za-z0-9+/]{3}=?)$/;
```
SamHellawell, line 85: not sure about this regex, dont think itll catch all cases or maybe itll have false positives

the dock-app repo already uses `@polkadot/util-crypto` - there shouldnt be a need to have an RPC call. you should be able to use `isBase64` from that module just fine

maycon-mello: if we import `@polkadot/util-crypto` in the app it will try to load wasm-crypto in the js thread (with js fallback instead of web-assembly), which can create big performance issues in react-native

maycon-mello: https://github.com/docknetwork/react-native-sdk/pull/107 i've added a PR in the wallet-sdk repo

maycon-mello: image

src/features/credentials/hooks/usePrettyVCData.ts
```diff
@@ -80,23 +81,62 @@ const applyCustomDateFormat = ({vcData}: CustomMappingProps): VCData => {
   return vcData;
 };
 
+const MAX_ATTR_LENGTH = 128;
+
+const base64Regex =
+  /^(?:[A-Za-z0-9+/]{4})*?(?:[A-Za-z0-9+/]{2}(?:==)?|[A-Za-z0-9+/]{3}=?)$/;
+// TODO: use wallet-sdk base64 util
+// Will need json-rpc method for that
+function isBase64(value: string) {
```
SamHellawell, line 90: need to remove this method+regex

maycon-mello: fixed

SamHellawell (dismissed):

SamHellawell (approved):

PR #509, opened 2023-03-17
fix/relay service background mode

maycon-mello:


src/features/relay-service/message-handler.js
```diff
@@ -84,9 +84,14 @@ export function handleVerificationResult({message, sendReply, data}) {
 // TODO: Move this to the wallet-sdk
 // Will need to support other message types
 function handleVCDistribution(data, rawMessage) {
-  debug('vc distribution received', data.credentials);
+  debug('vc distribution received', JSON.stringify(data));
```
mike-parkhill, line 87: we probably want to be careful about this - it will contain PII

src/features/relay-service/use-relay-service-data-fetcher.test.ts
```diff
@@ -12,6 +13,7 @@ describe('useRelayServiceDataFetcher', () => {
   let getMessagesSpy: jest.SpyInstance;
 
   beforeEach(() => {
+    AppState.currentState = 'active';
```
mike-parkhill, line 16: should probably have a test to ensure we don't fetch messages in other states

src/features/relay-service/message-handler.js
maycon-mello: fixed

src/features/relay-service/use-relay-service-data-fetcher.test.ts
maycon-mello: done, test added

mike-parkhill (approved):

PR #512, opened 2023-03-28
fix/pex filters

maycon-mello:
dev


maycon-mello:
Depends on https://github.com/docknetwork/react-native-sdk/pull/108

PR #513, opened 2023-03-30
feat/filter credentials by issuer and holder

maycon-mello:
Filter by holder and issuer DID


src/features/credentials/filters/DatePicker.tsx
```diff
@@ -0,0 +1,55 @@
...
+        from: filter.value?.from || new Date(),
+        to: filter.value?.to || new Date(),
+        [key]: timestamp ? new Date(timestamp) : new Date(),
+      },
+    });
+  };
+
+  return (
+    <Box mb={2}>
+      <FormControl w="100%" isInvalid px={5} mt={2}>
```
mike-parkhill, line 31: do we need to worry about date formats in different locales causing us issues?

src/features/credentials/filters/CredentialFilters.tsx
```diff
@@ -105,6 +117,31 @@ function FilterSelector({
     );
   }
 
+  if (filterOption?.type === 'text') {
```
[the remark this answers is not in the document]
maycon-mello: fixed

src/features/credentials/filters/DatePicker.tsx
maycon-mello: The date format will be handled by the native component. We get back a regular js Date object, which is used to filter by timestamp. This approach works for any locale

mike-parkhill (approved):

PR #514, opened 2023-03-31
feat/filter by expiration and sorting

maycon-mello:
- Filter by expiration date
- Sorting


src/features/credentials/filters/components/FilterExpirationDate.tsx
```diff
@@ -0,0 +1,32 @@
...
+          });
+        }}
+        placeholder={'Select a value'}
+        _selectedItem={{
+          bg: 'teal.600',
+          endIcon: <CheckIcon size={5} />,
+        }}
+        mt="1">
+        <Select.Item key={0} label={'Select a value'} value={null} />
+        <Select.Item key={1} label={'Hide expired'} value={'notExpired'} />
```
mike-parkhill, line 27: I was expecting this to be a date range filter (i.e. show creds that expired between date 1 and date 2), but this is probably a good start

we can add the date range later if people actually ask for it

maycon-mello: Sounds good. I've added support for date range in the filtering function, would need to update the UI only.

maycon-mello: just added a unit tests for that

src/features/credentials/filters/components/SortingDropdown.tsx
```diff
@@ -0,0 +1,89 @@
+import {SortingDirection, SortingOptionID} from '../filter-types';
+import {Box, CheckIcon, FormControl, Select} from 'native-base';
+import React from 'react';
+import {translate} from '../../../../locales';
+
+const SORTING_OPTIONS: SortingOptionID[] = [
+  'issuanceDate',
+  'issuer',
+  'expirationDate',
+  'holder',
```
mike-parkhill, line 10: should we display the holder/subject ID on the preview card so this sort makes more sense?

maybe issuer name or DID, too?

maycon-mello: I agree, can we add this as part of https://dock-team.atlassian.net/browse/CM-964?

src/features/credentials/filters/components/SortingDropdown.tsx
```diff
@@ -0,0 +1,89 @@
...
+  onChange,
+  filter,
+}: {
+  onChange: (filter: any) => void;
+  filter: any;
+}) {
+  return (
+    <Box mb={2}>
+      <FormControl w="100%" isInvalid px={5}>
+        <FormControl.Label>Sort by</FormControl.Label>
```
mike-parkhill, line 24: are you still waiting on UX designs before moving these labels to translation files?

maycon-mello: Yes

src/features/credentials/filters/credential-filters.tsx
```diff
@@ -108,3 +113,133 @@ export function filterByText({credentials, filter}: FilterFnInput) {
...
+
+      if (direction === 'asc') {
+        return aDate - bDate;
+      }
+
+      return bDate - aDate;
+    });
+  }
+
+  if (id === 'expirationDate') {
```
mike-parkhill, line 203: similar to the control map for the filters, would be nice to make the sort rules pluggable too.

src/features/credentials/filters/use-credential-filters.test.ts
```diff
@@ -171,4 +183,269 @@ describe('useCredentialFilters', () => {
...
+        }) as any,
+        createMockCredential({
+          expirationDate: moment().add(10, 'days').toDate().toString(),
+        }) as any,
+        createMockCredential({
+          expirationDate: undefined,
+        }) as any,
+      ];
+    });
+    it('expect to not return expired credentials', () => {
```
mike-parkhill, line 203: dateRange test?

Do you have a test where no expiration filter is specified but the sample set has both expired and non-expired creds so we can verify that all creds are returned regardless of expiration?

maycon-mello: Done, test was added

src/features/credentials/filters/credential-filters.tsx
```diff
@@ -108,3 +113,133 @@ export function filterByText({credentials, filter}: FilterFnInput) {
...
+    return credentials;
+  }
+
+  const filterInput: ExpirationDateRange = filter.value as ExpirationDateRange;
+
+  if (!filterInput) {
+    return credentials;
+  }
+
+  if (filterInput.mode === 'notExpired') {
```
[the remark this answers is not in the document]
maycon-mello: This is meant to be an alias. The string type has better reusability allowing us to added new modes quickly, such as: 'notExpired' | 'expired' | 'expiring' | 'dateRange'. 

Maybe I can rename `notExpired` to `valid`

src/features/credentials/filters/credential-filters.tsx
maycon-mello: fixed

src/features/credentials/filters/components/SortingDropdown.tsx
maycon-mello: done

src/features/credentials/components/CredentialCard/CredentialCard.tsx
```diff
@@ -128,6 +132,38 @@ export function CredentialCard({
...
+                fontSize={'13px'}
+                fontWeight={600}
+                fontFamily={Theme.fontFamily.montserrat}>
+                Issuer:
+              </Typography>
+              <Typography
+                fontSize={'13px'}
+                fontWeight={300}
+                fontFamily={Theme.fontFamily.montserrat}>
+                {credential?.issuer?.id}
```
mike-parkhill, line 147: name might be more meaningful here

mike-parkhill: if we have it...

so maybe issuer?.name || issuer?.id or something

src/features/credentials/components/CredentialCard/CredentialCard.tsx
```diff
@@ -128,6 +132,38 @@ export function CredentialCard({
...
+                fontSize={'13px'}
+                fontWeight={600}
+                fontFamily={Theme.fontFamily.montserrat}>
+                Holder:
+              </Typography>
+              <Typography
+                fontSize={'13px'}
+                fontWeight={300}
+                fontFamily={Theme.fontFamily.montserrat}>
+                {credential?.credentialSubject?.id}
```
mike-parkhill, line 163: again, probably better UX to show the name if it exists

mike-parkhill:
These implementations will have to change a fair bit when we settle on a different data store to optimize for larger credential sets, but for now this approach will work.

src/features/credentials/components/CredentialCard/CredentialCard.tsx
maycon-mello: fixed

src/features/credentials/components/CredentialCard/CredentialCard.tsx
maycon-mello: fixed

maycon-mello:
> These implementations will have to change a fair bit when we settle on a different data store to optimize for larger credential sets, but for now this approach will work.

Yep, I was planning on using typeorm (sqlite). We can still using the same structure for filters, but instead of passing a credential list as parameter for each filter, we could pass a queryBuilder.  something like: 

```js
queryBuilder.where('credential.type = :type', { type: 'BasicCredential' });
queryBuilder.andWhere('credential.issuanceDate > :date', { date:'some-date' });

```

I think unit tests would keep the same

mike-parkhill (dismissed):

mike-parkhill (approved):

PR #516, opened 2023-04-01
fix/date range filter timestamp

maycon-mello:


src/features/credentials/credential-filtering/CredentialFilters.tsx
```diff
@@ -27,7 +27,7 @@ export function CredentialFilters(props: CredentialFiltersProps) {
             filter={filter}
             facets={controller.facets}
             onChange={controller.onFilterChange}
-            canRemove={index > 0}
+            canRemove={true}
```
mike-parkhill, line 30: there's never a situation where we don't want them to be able to remove it?

If that's the case, do we need this attribute at all?

src/features/credentials/credential-filtering/filters/filter-by-issuance-date.ts
```diff
@@ -11,14 +26,17 @@ export function filterByIssuanceDate({credentials, filter}: FilterFnInput) {
     return credentials;
   }
 
+  const fromTimestamp = dateRange.from && toStartOfDayTimestamp(dateRange.from);
```
mike-parkhill, line 29: could these functions be shared between the two date filters? They're basically the same.

src/features/credentials/credential-filtering/CredentialFilters.tsx
maycon-mello: Yep, I think we can remove this. let me update the code

mike-parkhill (approved):

PR #521, opened 2023-04-04
Make Credentials as default screen

Jayteekay:
## Proposed Changes
- Rearranging the bottom navigation tabs in the order Credials, DIDs, QR Scanner, Tokens, and Settings
- Setting "Credentials" as the default screen when user is authenticated

## Testing instructions
- Before and after configurations of the bottom navigation can be seen below:
Screenshot 2023-04-04 at 10 36 58 AM -> Screenshot 2023-04-04 at 10 36 00 AM
- On login into the wallet app, the first screen shown should be the Credentials screen.


maycon-mello (approved):
looks good

mike-parkhill (commented):
Looks good, but wondering if it would make more sense to have Creds and DIDs to the left of the Scanner option and Tokens and Settings to the right. 

Feels like better logical grouping to me, but it's not a strongly held opinion.

maycon-mello:
> Looks good, but wondering if it would make more sense to have Creds and DIDs to the left of the Scanner option and Tokens and Settings to the right.
> 
> Feels like better logical grouping to me, but it's not a strongly held opinion.

good point, I agree. Creds, DIDs, QR Scanner, Tokens, and Settings looks a logical grouping

SamHellawell:
+1 on moving tokens to the right

Jayteekay:
> Looks good, but wondering if it would make more sense to have Creds and DIDs to the left of the Scanner option and Tokens and Settings to the right.
> 
> Feels like better logical grouping to me, but it's not a strongly held opinion.

Agreed. That change has been included now.

mike-parkhill (commented):
We don't have any snapshots that capture this?

Jayteekay:
> We don't have any snapshots that capture this?

maycon-mello (approved):

PR #523, opened 2023-04-06
fix/credential status check

maycon-mello:
- fix: issues with credential status check


mike-parkhill (commented):
Are you able to add tests that actually check the credential status?

maycon-mello:
> Are you able to add tests that actually check the credential status?

It would require an integration test actually, removing the mocks placed for the wallet-sdk. We have jira tickets for that already:
https://dock-team.atlassian.net/browse/DCKA-1236
https://dock-team.atlassian.net/browse/DCKA-963

Since all the wallet-sdk calls are being mocked, we can't catch jsonld issues for example. That's another reason to remove such mocks

mike-parkhill:
Let's get that prioritized

mike-parkhill (approved):

PR #526, opened 2023-04-10
DCKA 1477 - Dock wallet improvements on verification flow

Jayteekay:
## Proposed Changes
Implementation of some improvements to the Credential Presentation flow including:
- Making credentials selection checkbox respond to selection when any part of the card is clicked
- UI should differentiate between requested fields when sharing credential attributes
- Select a user's DID by default in the Identity selection stage. Select the first DID if the user has multiple DIDs.

## Testing Instructions
- To test the credential selection check box, 
   - Scan a credential verification request qr code
   - On the "Select the credentials to be presented" screen, select a credential by touching a credential card
   - Regardless of whether the checkbox icon is pressed, the card should be selected if any section of the card is pressed.
-  The UI changes made to differentiate requested fields from other fields is shown in the comparison below:
Screenshot 2023-04-10 at 12 13 08 PM -> Screenshot 2023-04-10 at 12 11 16 PM
- To test the default DID selection, On the "Select Identity to present" screen, the user's DID should be selected by default.

## Additional Notes
- For the issue "When you have two credentials of the same type, the only way to tell them apart is by the issuance date - maybe tap to preview?", the changes made to the checkbox in this PR make it tricky to implement at the preview of the credentials. I think we could show credential ID to differentiate credentials instead. Are there any other suggestions for this?
- I am unable to reproduce the error “Credential not verified: a DID document lookup was successful, but the DID in question does not exist. This is different from a network error.” I'd love it if I can get a reproducible set of steps.


src/features/credentials/SelectBBSHolderDIDToShareScreen.js
```diff
@@ -67,15 +72,19 @@ export function SelectBBSHolderDIDToShareScreenContainer({route}) {
     presentationDefinition,
   });
   const {dids} = useDIDAuth();
-  const [selectedDID, onSelectDID] = useState('');
+  const defaultDid = useMemo(
+    () => (dids?.length ? dids[0] : undefined),
```
mike-parkhill, line 76: Not part of this ticket I don't think, but another little improvement we could make is to allow the user to specify which DID should be the "default" on the DID management screen.

src/features/credentials/components/SelectDIDComponent.tsx
```diff
@@ -1,11 +1,22 @@
...
+import {DID} from '../types';
+
+interface SelectDIDComponentProps {
+  dids: DID[];
+  defaultDid: DID;
+  handleChange: (value: DID['value']) => void;
+}
 
-export function SelectDIDComponent({dids, handleChange}) {
+export function SelectDIDComponent({
```
mike-parkhill, line 15: do/can we skip this step completely if the user only has a single DID?

src/features/credentials/SelectBBSHolderDIDToShareScreen.js
Jayteekay: We could explore this. I just wonder how we could seamlessly integrate that into the UI. 

One option would be adding a "Make default" button. Another that comes to mind is to use the most recently used selection as default.

src/features/credentials/components/SelectDIDComponent.tsx
Jayteekay: By "skip", do you mean not showing the DID selection stage? If so, it means we assume that the user intends to use their single DID every time. 
It would turn out to be a bad UX for a user who wants to create a separate identity for the particular presentation, especially if they mistakenly omit the DID creation step before proceeding.

mike-parkhill: Do we allow them to create a new DID at that step? I didn't think we did

Jayteekay: No, we don't. Just pointing out that they have the option to confirm their DID before proceeding.

mike-parkhill: My assumption is that most users won't know their DID and that if there's only one DID on the device that there isn't much too confirm even if they do know it.

Obviously, if they have multiple DIDs then we need to let them confirm which one to use

Jayteekay: True. That makes sense. I'll make that change: Skip DID selection if the user has only a single DID.

src/features/credentials/SelectBBSHolderDIDToShareScreen.js
mike-parkhill: yeah, can you add a ticket to Jira so we can talk about that and come up with the plan?

mike-parkhill (approved):

PR #529, opened 2023-04-17
DCKA 1555 - WIP improve usage metrics

Jayteekay:
## Proposed Changes
- Implementation of descriptive screen names in analytics
- Implementation of additional metrics for events (import, export, create, delete, update) involving:
   - Credentials
   - DIDs
   - Tokens
   - QR scans
   - Settings
   - Dev Settings
   - Presentation
   - Account Backup
   
## Testing instructions
-  The debug analytics can be done on android with the following steps:
   - Run app on android
   - Run ```adb shell setprop debug.firebase.analytics.app com.dockapp```
   - On the analytics dashboard, Go to Admin -> Debug View to monitor the metric events.


src/features/qr-code-scanner/qr-code.js
```diff
@@ -151,6 +160,9 @@ export async function credentialHandler(data) {
       cancelText: translate('navigation.cancel'),
       onConfirm: async () => {
         await credentials.add(credentialData);
+        logAnalyticsEvent(ANALYTICS_EVENT.CREDENTIALS.IMPORT, {
```
mike-parkhill, line 163: are we able to log some details about the VC?
*  what signing algorithm was used in the proof (e.g. bbsplus, ed25519 , etc.)?
* credential format (json-ld vs jwt-vc, etc.)
*  credential type  (e.g. University Degree, Proof of Attendance, etc.)

src/features/analytics/analytics-slice.js
```diff
@@ -12,13 +12,14 @@ export function logAnalyticsEvent(eventName, params = {}) {
...
+    RECEIVE: 'token_receive',
+    BUY: 'token_buy',
   },
   WALLET: {
     BACKUP: 'wallet_backup',
     IMPORT: 'wallet_import',
+    DELETE: 'wallet_delete',
   },
   ACCOUNT: {
     BACKUP: 'account_backup',
```
mike-parkhill, line 25: what about account_create?

src/features/qr-code-scanner/qr-code.js
Jayteekay: Updated this.

src/features/analytics/analytics-slice.js
Jayteekay: I've added this now.

Jayteekay:
> is relay-service tracking still to come?

Yes. I'm still figuring this out.

src/features/qr-code-scanner/qr-code.js
```diff
@@ -139,6 +139,12 @@ export async function credentialHandler(data) {
       status = await getCredentialStatus(credentialData);
       if (status === CREDENTIAL_STATUS.VERIFIED) {
         await credentials.add(credentialData);
+        logAnalyticsEvent(ANALYTICS_EVENT.CREDENTIALS.IMPORT, {
+          status: 'success',
+          format: 'json-ld', // should be made dynamic when jwt-vc is supported.
+          signingAlgorithm: credentialData.proof.type,
```
mike-parkhill, line 145: we just saw in PolygonID credentials that `proof` is an array rather than object so should probably be a bit defensive here and at a minimum use null operators

Jayteekay: I've added a null check for this. Whenever we support PolygonID, I think we would have to make the change throughout the app to account for `proof` being an array.

mike-parkhill: Yeah, Maycon and I were messing about trying to import one of their credentials and he was able to make a small tweak to get it to import, but we didn't try verifying it or anything. If/when we get to proper support I'm sure there will be a fair amount of work but it was good to see what others up to and helps reinforce that we need to be defensive and not make too many assumptions that are only true for our VC's.

mike-parkhill: and I suppose any logging/telemetry should be very defensive - last thing you want is logging/telemetry breaking the app

Jayteekay:
> Are you still looking to add relay-service events? Or will that be a different PR?

I initially intended it to be in a separate PR. However, I will be able to add it to this one now that I've figured it out.

mike-parkhill (dismissed):

mike-parkhill (approved):

PR #530, opened 2023-04-17
fix/nested credential subject

maycon-mello:
Fixing issues with nested credential subjects on bbs+ verification


maycon-mello:
Will push unit tests on separate PR. need some refactoring on the code first

src/features/credentials/hooks/credentialPresentation.js
```diff
@@ -290,7 +290,7 @@ export function useBBSCredentialPresentation({presentationDefinition}) {
         if (isBBSPlusCredential(credential)) {
           const attributesToReveal = Object.keys(revealedAttributes[id]).map(
             attr => {
-              return `credentialSubject.${attr}`;
+              return `credentialSubject.${attr}`.replace(' ', '.');
```
mike-parkhill, line 293: where do the spaces come from in the first place? In the VC they wouldn't be there

maycon-mello: this seems related to pretty vc formatting, would need to use the rawCredential when selecting the keys, otherwise it could create other issues. Let me double check

maycon-mello: I would stick with this hotfix for now. We might need to update pretty vc to not change the attribute id, adding those spaces in the `property` or get this data from another attribute
image

can discuss with @SamHellawell

mike-parkhill: Ok, let's get a ticket to look at it.

Is it something to look at while doing the storage update?

mike-parkhill (approved):

src/features/credentials/hooks/credentialPresentation.js
maycon-mello: thanks, I will refactor this logic in the storage update. Also add better unit tests to cover all the scenarios we need

maycon-mello: Ticket created https://dock-team.atlassian.net/browse/DCKA-1610

PR #538, opened 2023-04-27
feat: handle didcomm qr codes

maycon-mello:


maycon-mello:
Would be good to add some tests on it, maybe we can push as separate PR

maycon-mello:
Relates to https://github.com/docknetwork/react-native-sdk/pull/113

maycon-mello:
image
image

maycon-mello:
@Jayteekay will add some unit tests to it

SamHellawell:
so does this go like QR -> notifications -> view instead of QR -> accept/decline -> view? guessing we dont have that state ready just yet, but that would be the ideal flow

also, are you submitting the ACK response to the relay after scanning and import was successful? we dont want the QRs to be scannable twice

maycon-mello:
> 



> so does this go like QR -> notifications -> view instead of QR -> accept/decline -> view? guessing we dont have that state ready just yet, but that would be the ideal flow
> 
> also, are you submitting the ACK response to the relay after scanning and import was successful? we dont want the QRs to be scannable twice

`accept/decline`  is not implemented. let me create a jira ticket for that.
`submitting the ACK response to the relay after scanning`, this part is missing too. let me create another follow up jira for this one too

SamHellawell:
>this part is missing too. let me create another follow up jira for this one too

itll take a few minutes to add a call to ACK, probably best just to do here. we dont want unencrypted messages containing VCs lingering longer than they have to

maycon-mello:
> > this part is missing too. let me create another follow up jira for this one too
> 
> itll take a few minutes to add a call to ACK, probably best just to do here. we dont want unencrypted messages containing VCs lingering longer than they have to

sure, will need to add a new method on the relay service rpc service. we can do on this PR

src/features/qr-code-scanner/qr-code.js
```diff
@@ -399,7 +402,38 @@ export async function relayServiceHandler(json) {
...
+  await relayServiceMessageHandler({
+    message,
+    callback: navigateBack,
+  });
+
+  const messageId = data.match(
+    /(?<=https:\/\/relay\.dock\.io\/read\/)[a-f0-9]+/,
+  )?.[0];
+
+  if (messageId) {
```
SamHellawell, line 425: at this point has the message been fully processed and converted into an imported credential?

SamHellawell (approved):
lgtm if the quick question i posted is a "yes" answer - which i think it is

PR #539, opened 2023-04-28
fix: fetch white label team id from configs

maycon-mello:


.github/workflows/white-label.yml
```diff
@@ -175,11 +176,8 @@ jobs:
           MATCH_PASSWORD: ${{ env.MATCH_PASSWORD }}
           FASTLANE_APP_IDENTIFIER: ${{ env.IOS_BUNDLE_ID }}
           FASTLANE_PASSWORD: ${{ env.MATCH_PASSWORD }}
-          FASTLANE_SESSION: ${{ secrets.FASTLANE_SESSION }}
-          FASTLANEAppStore__APPLE_APPLICATION_SPECIFIC_PASSWORD:  ${{ secrets.FASTLANE_APPLE_APPLICATION_SPECIFIC_PASSWORD }}
-          FASTLANE_APPLE_ID: ${{ secrets.FASTLANE_APPLE_ID }}
```
maycon-mello, line 180: will try to remove FASTLANE_APPLE_ID, this should not be required

maycon-mello: should be needed only when uploading to testflight from CI

mike-parkhill (approved):

PR #543, opened 2023-05-04
Ci/split white label builds

mike-parkhill:


.github/workflows/white-label-builder.yml
```diff
@@ -0,0 +1,49 @@
+name: Build White Label Deployments
+on:
+  workflow_dispatch:
+
+  pull_request:
+    branches:
+      - fix/white-label-team-id
```
maycon-mello, line 7: we should remove this before merging to dev right?

mike-parkhill: yep

mike-parkhill: we'll need to make sure this whole trigger section is returned to its original state

maycon-mello (approved):

PR #561, opened 2023-05-18
invalid credential toast displayed after scanning presentation request 

maycon-mello:


src/features/qr-code-scanner/qr-code.js
```diff
@@ -113,6 +113,12 @@ export async function credentialHandler(data) {
       throw err;
     }
 
+    try {
+      validateCredential(credentialData);
+    } catch (err) {
+      return;
```
SamHellawell, line 119: why are we now silently swallonig the error? what if the QR is an import credential QR but the credential is invalid? ideally we should check if the data IS a VC rather than if its just valid

maycon-mello: I think the method name is not good. This method will only check if the data has the attributes a VC would require, so its not actually doing a data validation, which will be performed in a next step

src/features/qr-code-scanner/qr-code.js
```diff
@@ -215,9 +239,18 @@ export function getMappedErrorMessage(message) {
 }
 
 export async function onPresentationDefinitionScanned(value) {
-  const rawData = isValidUrl(value)
-    ? await getDataFromUrl(value)
-    : stringToJSON(value);
+  let rawData;
+
+  if (isValidUrl(value)) {
```
SamHellawell, line 244: ideally you dont want to be fetching the data here and showing toasts, its prone to breakage if the order changes. its also forcing every QR handler to load the data individually, which can be slow or even expire and fail 2nd request

id advise to refactor so that:
- initial qr handler (at point of scanning, before calling this list) resolves the URL if its an URL (follows redirects if any to the final one if http) and fetches the data, then passes around the resolved URL and data - or original scanned content and null data if not an url

src/features/qr-code-scanner/qr-code-helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+  );
+}
+
+function isCertsCredentialUrl(url) {
+  const isCerts = isCertsURL(url);
+  const blacklist = [
+    '/proof/', // proof requests
+  ];
+
+  return isCerts && !blacklist.some(item => url.includes(item));
```
SamHellawell, line 22: string include checking is not a good way to be sure an url is not what you want. take for example:
```
https://creds.dock.io/credential_id?utm_source=blacklistedparam
```

at the very least strip off anything before `?` for the check, or use a proper solution to check the pathname excluded from other URL parts. should be easy with https://developer.mozilla.org/en-US/docs/Web/API/URL

SamHellawell: similarly we can update url.startsWith to just check if hostname === which should be quicker and more robust (minor difference not really too fussed on that one)

src/features/qr-code-scanner/qr-code-helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+export async function getQRCodeHandlerContext(qrData) {
+  let context = {
+    rawData: qrData,
+    jsonData: null,
+  };
+
+  const isUrl = isValidUrl(qrData);
+
+  if (isUrl) {
+    context.jsonData = await handleURL(qrData);
```
SamHellawell, line 54: `handleURL` should return the changed URL string incase of redirect following aswell as data. you should also pass that around or original url in jsondata as it could potentially be needed by a handler. eg: `openid://https://blah.com/issuer`

src/features/qr-code-scanner/qr-code-helpers.ts
```diff
@@ -0,0 +1,60 @@
+import {isValidUrl, stringToJSON} from '../../core/storage-utils';
+import {getDataFromUrl} from '../didManagement/didManagment-slice';
```
SamHellawell, line 2: nitpick: `getDataFromUrl` shouldnt exist in `didManagement` (doesnt have to be solved here)

src/features/qr-code-scanner/qr-code-helpers.ts
maycon-mello: thanks, I've pushed some updates

src/features/qr-code-scanner/qr-code-helpers.ts
maycon-mello: yeah, I've fixed that too

SamHellawell (dismissed):

SamHellawell:
lgtm i think but will need test fixing + decent QA

SamHellawell (approved):

